Issue #60, opened 2020-10-09
Auto-scale option or headless deposits


actuallymentor:
Right now if a node wants to be maximally useful by offering a maximum number of minipools they need to:

1. Check if refunds are available (can be automated now with `grep`)
2. Run the refund command (can be automated now with piping)
3. Run the deposit command (can *not* be fully automated easily)

Both do not have a headless option. The first can be automated with piping since it is a one step process, but the second does not.

Ideally there should be a config option for autoscaling.

If that is too much work then a `rocketpool deposit 16 -y` that deposits 1 and accepts the network rate should suffice. Node operators can then check the refund status programatically and trigger the above command.

moles1:
We discussed this concept internally but decided against implementing it for now due to the desired UX flow:

- We wanted node operators to perform refunds manually/explicitly to ensure that they know they are there and they can use the refunded ETH accordingly. If refunds are automated, some operators may be unaware that they are occurring and that the refunded ETH can be re-staked. (Note that this could have been included in the `node` daemon but was implemented as explicit API calls instead).

- We wanted deposits to only be performed manually due to the fluctuations in network commission rate. It seemed presumptuous that auto-scaling minipools would be desirable for node operators given that the commission rate could potentially drop and saddle them with a large number of minipools earning a low fee.

On the second point, the commission rate mechanics may change according to RPIP-003, so it's definitely worth consideration if that happens.

I'll leave this issue open to gauge whether others would like to see these features implemented - it seems likely that they should make it into a future release at some point :)

actuallymentor:
Thanks for the thoughtful response!

> We wanted node operators to perform refunds manually/explicitly to ensure that they know they are there and they can use the refunded ETH accordingly.

Perhaps an automated option through the CLI for those of us who use more advanced scripting? I have a cron that checks resource utilisation and makes a calc to see if I want to withdraw and potentially restake.

> We wanted deposits to only be performed manually due to the fluctuations in network commission rate.

Similar response to the above. Maybe don't make it default, but the automation addicts like myself would love to at least be able to, even if it takes more work to set up.

moles1:
All of the above sound like good additions. We've got some more urgent issues to tackle for now, but should be able to circle back to this at some point :)

moles1:
Implemented (along with other non-interactive option flags) on branch https://github.com/rocket-pool/smartnode/compare/v0.0.7

actuallymentor:
[...]

PR #72, opened 2020-10-29
Issue #60 command line options for deposit

kidkal:
Add command line options for the `rocketpool node deposit` command.  Help message looks like:
```
NAME:
   rocketpool node deposit - Make a deposit and create a minipool

USAGE:
   rocketpool node deposit [options]

OPTIONS:
   --amount value, -n value  Amount of ETH to deposit (0|16|32) (default: 0)
   --autofee, -a             Use the suggested fee rate
   --fee value, -f value     Minimum node commission rate in % [0-20] (default: 0)
```
**Implementation notes:**
These command line options allows non-interactive entry to this command.  Error handling and range checks are from existing logic for interactive input.  the `--autofee` option will override the `--fee` option if they are set together.

### DIFF
--- a/rocketpool-cli/node/commands.go
+++ b/rocketpool-cli/node/commands.go
@@ -67,7 +67,21 @@ func RegisterCommands(app *cli.App, name string, aliases []string) {
                 Name:      "deposit",
                 Aliases:   []string{"d"},
                 Usage:     "Make a deposit and create a minipool",
-                UsageText: "rocketpool node deposit amount",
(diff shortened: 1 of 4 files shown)


moles1:
awesome, thanks for this :) will review soon and merge it in after making any tweaks!

kidkal:
did another commit to push the "suggested node fee" calculation into the service.  This makes the CLI client a bit neater.  If you don't like this because it messes with the api types `NodeFeeResponse` (since it's also used in other places) I can revert.

Issue #85, opened 2021-01-28
Watchtower nodes not reporting withdrawn minipool balances


moles1:
Currently, when a minipool is withdrawn from it is destroyed and removed from the `RocketMinipoolManager` active minipool index. This means that watchtower nodes don't account for withdrawn minipools when calculating total network balances.

A new index should be created for "active and withdrawn" minipools, which they are not removed from until their withdrawal is processed.

The watchtower network balance calculations should then be updated to query the "active and withdrawn" minipool index. The watchtower should query `getMinipoolWithdrawalTotalBalance` and `getMinipoolWithdrawalNodeBalance` for minipool balance data if a minipool is already withdrawn. Otherwise, it should query the minipool contract itself and perform calculations as usual.

moles1:
This issue is now resolved as minipool contracts are no longer destroyed until their validator balance has been received and distributed to the nETH contract and rETH contract/deposit pool.

PR #91, opened 2021-02-15
Fix withdrawn minipool balances

kidkal:
Fixes rocket-pool/smartnode#85
- use the `minipools.unprocessed.index` collection of minipools
- calculate the withdrawable balance to the staking pool
- indicate that the minipool has been withdrawn but not yet processed
- add withdrawn but unprocessed staking balance to the tally


moles1:
Apologies, these changes are now redundant due to a number of changes to the RP contracts. The "unprocessed" minipool index has been removed, and trusted nodes are no longer responsible for processing validator withdrawals into the network.

Issue #80, opened 2020-12-08
Prysm watchtower nodes reporting incorrect network balance


moles1:
Prysm watchtower nodes report an incorrect network balance. This appears to be related to the validator status/balance request.

PR #93, opened 2021-02-18
Fix Prysm watchtower network balance

kidkal:
Fixes rocket-pool/smartnode#80
- Prysm api response to ListValidators and ListValidatorBalances returns paged responses, so we need to make multiple requests with supplied NextPageToken until there are no more pages
- Make Lighthouse validator requests in batches as well and not iterate through the entire set of validators


moles1:
Nice catch on the pagination for the Prysm results, and good call on batching the LH validator requests rather than querying the whole validator set, that looks much better :)

Issue #78, opened 2020-11-30
Add gas price & limit options


moles1:
- Add gas price & gas limit flags to CLI endpoints which send transactions
- Add gas price & gas limit config options for node and smartnode daemon tasks

moles1:
Merged in!

PR #94, opened 2021-02-20
Ability to set gas price and limit on transactions

kidkal:
Fixes rocket-pool/smartnode#78

- add gas properties to RocketPoolConfig so it can be pulled in via settings and config yml files
- add gas optional parameters to command line
- pass gas option parameters through to api service
- save gas parameters to Wallet and set to TransactOpts when GetNodeAccountTransactor is called

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,3 +32,6 @@ require (
 	google.golang.org/grpc v1.29.1
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+// TODO remove this and update with proper rocket-pool/rocketpool-go version after merge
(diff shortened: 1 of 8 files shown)


kidkal:
testing gas parameters with some fun numbers:
`rocketpool --gasPrice 4444444444 --gasLimit 99999 minipool refund`

then looking at https://goerli.etherscan.io/
image

kidkal:
The last commit depends on PR https://github.com/rocket-pool/rocketpool-go/pull/3. Once this is merged, it would be possible to easily add gas estimates for other rocketpool functions.

kidkal:
It's probably best to start a new pull request at this point since most of the original functionality has been merged.

PR #98, opened 2021-03-11
estimate gas cost before proceeding

kidkal:
example of command (node deposit) to estimate gas usage.  This uses new utility function in rocketpool-go project: `Contract.GetGasInfo` for a particular contract's method.  The rocketpool-cli will display the estimated gas cost, ask for confirmation before proceeding


kidkal:
this is redundant since #130

PR #122, opened 2021-03-31
Add validator index to minipool status

PengFei-eth:
I have no clue why visual studio code is adding spaces to the imports only. Any advice?

### DIFF
--- a/shared/services/beacon/client.go
+++ b/shared/services/beacon/client.go
@@ -1,10 +1,9 @@
 package beacon
 
 import (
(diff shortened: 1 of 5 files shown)


moles1:
Hey mate, thanks for this :) it looks like in this PR you're querying the validator index separately within each API client, but it's actually already in the validator responses they return internally so it doesn't need to be fetched again. I've implemented this functionality in https://github.com/rocket-pool/smartnode/commit/ba4789ac1bf91488d701f85b4d001df52b5f6dfd so will close this PR, cheers though!

PR #128, opened 2021-04-08
Add More Settings to the Service Config Interview

jclapis:
This is one of two PRs that adds some additional parameters to the `service config` interview process, the other being https://github.com/rocket-pool/smartnode-install/pull/44. This specific PR adds support to the smartnode stack for parsing and validating user parameters based on their type. It currently handles `uint` and `uint16`, in addition to the standard `string` behavior.

This was originally meant to go to `beta-3`, so for `v2.5-Tokenomics`, you just need the last commit (`fbefdfd`).


moles1:
Thanks, pulled the relevant commit into `v2.5-Tokenomics` :)

Issue #129, opened 2021-04-11
[FR] Let the CLI report TX details (gas price, total cost, TX ID) when submitting a TX


jclapis:
This is requst for an improvement to the CLI UX when the user invokes a command that submits a transaction to the eth network. The current UX has two problems:

1. The user doesn't get to see how much the TX costs based on their current gas price setting, so it can come as somewhat of a shock to run a transaction (e.g. creating a minipool) and learn that it just cost you 0.2 ETH on mainnet.
2. The user doesn't have any feedback after confirming the transaction; the CLI just hangs and waits for it to be mined. On mainnet with the wrong gas settings this can take *hours*.

The feature request here is to submit gas and transaction details, then either wait for the transaction to be mined or let the user exit after the transaction has been posted so they can do other things.

We envision that the UX could take two forms:

A
```
...
Are you sure you want to stake 160.000000 RPL? Staked RPL can only be withdrawn after a delay.
y

With your current gas price setting of 2 Gwei, this will cost up to 0.000513844 ETH in gas fees.
Do you want to continue? [y/n]
y

Transaction has been submitted with hash 0xb3680215cd509df88baf4fd89260e4c0b6bc676669325853dbd6413ec34e252c.
You may follow its progress by visiting:
https://goerli.etherscan.io/tx/0xb3680215cd509df88baf4fd89260e4c0b6bc676669325853dbd6413ec34e252c

Waiting for the transaction to be mined... you may wait here for it, or press q to exit and return to the terminal.

Successfully staked 160.000000 RPL.
```

B
```
Are you sure you want to stake 160.000000 RPL? Please note that:

  * Staked RPL can only be withdrawn after a delay.
  * With your current gas price setting of 2 Gwei, this will cost up to 0.000513844 ETH in gas fees.

Do you want to continue? [y/n]
y

Transaction has been submitted with hash 0xb3680215cd509df88baf4fd89260e4c0b6bc676669325853dbd6413ec34e252c.
You may follow its progress by visiting:
https://goerli.etherscan.io/tx/0xb3680215cd509df88baf4fd89260e4c0b6bc676669325853dbd6413ec34e252c

Waiting for the transaction to be mined... you may wait here for it, or press q to exit and return to the terminal.

Successfully staked 160.000000 RPL.
```

*([See here](https://discord.com/channels/405159462932971535/704214707904446535/830544837500469309) for the discussion.)*

Though opinions on which one is best differ, either case seems to be preferable to the current UX according to Discord.

[...]

PR #133, opened 2021-04-16
Add async support to TX CLI commands

jclapis:
This adds asynchronous support to each of the CLI commands that involve submitting transactions to eth1. It will report the transaction ID and provide the user with a link to etherscan so they can watch it manually if it's taking a long time, then block (and give the user the option to exit if they want to get back to the terminal).

This is part of a series - it relies on https://github.com/rocket-pool/rocketpool-go/pull/5.

This is a partial fix for #129 - it doesn't include gas estimation, that will come from @kidkal's work.

**NOTE: This affects a *ton* of commands, and it needs to be tested extremely thoroughly before deployment!**


shared/utils/cli/utils.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// Print a TX's details to the console.
+func PrintTransactionHash(hash common.Hash) {
+
+    hashString := hash.String()
+
+    fmt.Printf("Transaction has been submitted with hash %s.\n", hashString)
+    fmt.Printf("You may follow its progress by visiting:\n")
+    fmt.Printf("https://goerli.etherscan.io/tx/%s\n\n", hashString)
```
kidkal, line 16: shouldn't hard-code the url, either leave it out or put into config

shared/utils/cli/utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+
+// Print a TX's details to the console, but inform the user NOT to cancel it.
+func PrintTransactionHashNoCancel(hash common.Hash) {
+
+    hashString := hash.String()
+
+    fmt.Printf("Transaction has been submitted with hash %s.\n", hashString)
+    fmt.Printf("You may follow its progress by visiting:\n")
+    fmt.Printf("https://goerli.etherscan.io/tx/%s\n\n", hashString)
```
kidkal, line 29: ditto

shared/utils/cli/utils.go
jclapis: Agreed, I put it into the config: https://github.com/rocket-pool/smartnode-install/commit/ea2f26be32e6abc8e5cab34be3ac4b72397f5e69

PR #145, opened 2021-05-15
Change formatted API error response to use API response instead of a nil error

nii236:
Error messages from API calls were not formatted with the correct variable.

```
nii236 at ethstaker in ~ 
$ rocketpool odao member-settings

Could not get oracle DAO member settings: %!w(<nil>)
```

This pull request changes the `fmt.Errorf` from using `err`, which is the error from the HTTP client, to `Response.Error` which is the error message from the HTTP response.


jclapis:
Nice find, thank you for fixing this!

PR #148, opened 2021-05-18
Pocket support

jclapis:
This adds full Pocket compatibility to the Smartnode stack. It includes support for an incompatible client system (since Nimbus can't work with Pocket because of websockets) which removes incompatible ETH2 clients from the choice list once you've picked an ETH1 client. It also closes #132.


darrenlangley (approved):
LGTM

PR #171, opened 2021-12-09
Prompt for user input before printing sensitive data in wallet exports

### LINKED ISSUE #167: "rocketpool wallet export" command prints sensitive data to /dev/stdout without warning.
`rocketpool wallet export` should not print such critical data to stdout without warning the user first.

Perhaps without `--force` it should:
1) refuse to print to stdout until user input confirms the warning is acknowledged
2) exit immediately with an error if stdout is not a tty (avoiding issues with redirecting the output, when user input is required)

with --force it should behave as it does today.

_issue submitted by anonymous request_

jshufro:
Closes #167 

Y'all have some interesting code style for golang

NB- the printing is to stderr. Two things: urfave cli should default to stderr but appears to be printing to stdout. I can't figure out why. It does provide a package-level override, and it's possible some other import is overriding it. Additionally, most of the error reporting in this project appears to print to stdout. This is fine for most use cases, but since `wallet export` intuitively will be redirected to a regular file, the errors should be printed to stderr. This avoids situations where the regular file is created, but contains an error message, and the user thinks they have successfully exported the wallet when they actually have not.

I also noticed that when the app exits with an error, it has exit code 0. Not ideal, as i could imagine a bash script that wants to call something like `wallet export`, ignore stderr, and check the return code.

### DIFF
--- a/rocketpool-cli/wallet/commands.go
+++ b/rocketpool-cli/wallet/commands.go
@@ -1,9 +1,6 @@
 package wallet
 
 import (
-    "fmt"
-    "os"
-
     "github.com/urfave/cli"
 
     cliutils "github.com/rocket-pool/smartnode/shared/utils/cli"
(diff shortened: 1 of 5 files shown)


rocketpool-cli/wallet/commands.go
```diff
@@ -9,6 +12,19 @@ import (
 
 // Register commands
 func RegisterCommands(app *cli.App, name string, aliases []string) {
+    var stdoutIsTTY bool = false
```
jclapis, line 15: This is close - can you move all of this into `rocketpool-cli/wallet/export.go` like you've done with the stuff in `rocketpool-cli/wallet/init.go`? That should be where the checking logic occurs, not in the command registration bit. Thank you!

jclapis:
Looks good, thank you!

PR #185, opened 2022-01-12
Fix a couple staticcheck revealed issues

jshufro:
1- random client selection allows supermajority clients, even when
includeSupermajority is false.
2- one ignored Error

The random selection is a bit more complex now. I didn't want to open the possibility of an infinite loop if both `Fallback == true` and `includeSupermajority == false` and no clients fit the criteria. Instead, the loop iterates the client list once, collecting eligible clients into a slice, and erroring if resulting vector is empty. Otherwise, it randomly picks a client from the new slice.

N.B. This is untested, as I'm not currently set up on prater


jclapis:
Tested 100 times, and as expected it works great. Thank you!

PR #188, opened 2022-01-29
Use Start/Wait instead of Run to handle stdin/stdout better

SN9NV:
Fix for the `Error piping stdout: read |0: file already closed` bug.

From the [docs](https://pkg.go.dev/os/exec#Cmd.StderrPipe): 
> ... it is incorrect to use Run when using StderrPipe.

I tested it with a few of the standard commands and it appears to work as expected. I did not test with the SSH method. I see that has been marked as deprecated. I've never used it before, so I don't have that set up.

A kind note: It was really painful to contribute this change because `gofmt` is not being used for this project and I almost did not make a contribution because it was not used - knowing that my git stage would be a big mess. Please consider running `go fmt` over the code base. It is a standard in the industry. It will lower the barrier to entry for new contributes who have all their standard IDE setup to format on save. Thank you for this consideration.

### DIFF
--- a/shared/services/rocketpool/client.go
+++ b/shared/services/rocketpool/client.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	osUser "os/user"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/a8m/envsubst"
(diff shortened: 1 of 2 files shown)


SN9NV:
I would make `printOutput` a method of the `command` struct because it only uses the `command` variable of `Client`, so it makes more sense to be there. But that would have been a lot more refactoring. So I can make that another PR if it is what you would like.

jclapis:
I just merged this and formatted the repository as requested. Should be all set now, thank you!

SN9NV:
Fantastic. Thank you very much.

PR #190, opened 2022-02-17
Add missing cli parameter 'amount' for find-vanity-address

jshufro:


jclapis:
Looks good, thank you!

PR #192, opened 2022-02-21
Roughly double speed of vanity salt mining by applying trick from c6abf85... to address generation

jshufro:


jclapis:
Okey dokey, let me pull this in and give it a quick trial run. Thank you!

PR #193, opened 2022-02-26
Update wallet init mnemonic flow

jshufro:
5 discrete commits

1) Use sneaky unicode to prevent users from copy-pasting their mnemonic into the confirm dialog
2) Colorize the warnings about mnemonic security
3) Remove troubled --confirm-mnemonic flag
4) Update rocketpool_api to allow the smartnode to delete a wallet file. This command is not exposed outside the api.
5) Clear the terminal after the mnemonic is displayed, and before the user enters it to confirm. Allow the user to seamlessly start over by removing the old wallet if they fail the confirmation flow.

### DIFF
--- a/rocketpool-cli/wallet/utils.go
+++ b/rocketpool-cli/wallet/utils.go
@@ -11,7 +11,9 @@ import (
 	cliutils "github.com/rocket-pool/smartnode/shared/utils/cli"
 )
 
-var UTF8_EMPTY_SPACE string = "\xE2\x80\x8B"
+const UTF8_EMPTY_SPACE string = "\xE2\x80\x8B"
+const colorYellow string = "\033[33m"
+const colorReset string = "\033[0m"
 
 // Prompt for a wallet password
 func promptPassword() string {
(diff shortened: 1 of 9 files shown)


rocketpool/api/wallet/remove.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/rocket-pool/smartnode/shared/services"
+	"github.com/rocket-pool/smartnode/shared/types/api"
+)
+
+func removeWallet(c *cli.Context) (*api.RemoveWalletResponse, error) {
+
+	// Get services
+	if err := services.RequireNodePassword(c); err != nil {
+		return nil, err
+	}
```
jshufro, line 17: I can probably remove this. DIdn't dive too deep

rocketpool-cli/wallet/utils.go
```diff
@@ -20,35 +26,65 @@ func promptPassword() string {
...
+		fmt.Println("")
 	}
 }
 
 // Prompt for a recovery mnemonic phrase
 func promptMnemonic() string {
 	for {
 		mnemonic := cliutils.PromptPassword("Please enter your recovery mnemonic phrase:", "^.*$", "")
+		// If the user copy-pasted their mnemonic, don't make them panic by failing on the UTF8 guardians
+		mnemonic = strings.ReplaceAll(mnemonic, UTF8_EMPTY_SPACE, "")
```
jshufro, line 41: latently added these two lines.

it seems like a user should be allowed to copy paste their mnemonic on creation, as long as they can manually enter it, and then if they need to recover, we shouldn't punish them.

jshufro:
Superceded

PR #196, opened 2022-03-26
Print a warning when the externalip library returns ipv6

jshufro:
Nimbus's extip field will silently accept ipv6, but per the nimbus team, it is not currently supported.

From the `net` docs:
> To4 converts the IPv4 address ip to a 4-byte representation. If ip is not an IPv4 address, To4 returns nil.


jshufro:
cc @poupas

jclapis:
Thanks, and good to know it's not supported by Nimbus.

PR #197, opened 2022-04-04
DRY out the rocketpool service version string formatting code, and fix a copy paste victim

jshufro:
This may be too aggressively DRY (ie, the readability trade-off might not be worth it?)


jclapis:
I think it's fine, thanks for fixing this!

PR #204, opened 2022-05-09
add basic beaconcha.in node metrics support

InvisibleSymbol:
Only supports Lighthouse & Teku rn as I'm working on how to handle the exporter container for nimbus & prysm. Also has no first-setup screen yet.

Depends on https://github.com/rocket-pool/smartnode-install/pull/71 to fully work

Screenshot_20220508-233153
Screenshot_20220509-074826


jclapis:
So it begins... thanks Invis!

Issue #200, opened 2022-04-10
[FR] `rocketpool wallet recover` should prompt word-at-a-time, and validate the input against the wordlist.


jshufro:
It's pretty frustrating trying to type 24 words with spaces into a terminal that is masking the input, only to find a small typo means you must restart from the beginning.

Instead, just prompt the user for the length of their phrase, and then prompt N times for each word, validating as they go. If a word doesn't match the word list, re-prompt for that word.

Edit: While we're at it, we should allow input of _just_ the first 4 characters of each word from the wordlist, since that is enough to distinguish them, and many offline backup methods only retain 4 characters max per word (some words are only 3 characters long, eg, `ill`, and also need to work)

blinc117:
Also add for the test function

PR #205, opened 2022-05-12
Update mnemonic user input flow to be word-at-a-time and tolerant of partial words (minimum 4 letters)

jshufro:
Addresses https://github.com/rocket-pool/smartnode/issues/200

Tested on `wallet init` `wallet recover` and `wallet test-menmonic`


jclapis:
Thanks! Let's give this a shot.

PR #206, opened 2022-05-18
Correctly identify which containers are dirtied by EC settings when adjusting fallback

jshufro:


jclapis:
Good one, thanks! I'll tweak it a bit to make it work for fallback ECs as well.

PR #210, opened 2022-06-01
Add node 'ENABLE_METRICS=true' environment override

benv666:
In the #support channel on discord we had a discussion on enabling metrics for a standalone prometheus setup.
See https://github.com/rocket-pool/smartnode-install/issues/67 for how this used to work.

However, it seems that this method broke on the node part, since that no longer starts the metrics server when it's disabled in the configuration file. This patch adds an override check to see if the `ENABLE_METRICS` is set to `true`, and if so starts it anyway.

Alternatively this might be patched in through a separate configuration option, but I'll leave that to Joe :)


jclapis:
> hmm shouldn't the environment variables be opaquely referenced in the config abstraction layer?

Yes, they should, but I have to think harder about how to do that in a way that gives people fine-grained control over the metrics. I'll merge this for now so Ben can at least make it work properly, and robustify the metrics config in a later release.

PR #220, opened 2022-07-31
1058 - Adding proposer name to proposal details.

0xfornax:
Adding oDAO proposer name to `rp o o l`, `rp o o d`, and `rp o o v`.


jclapis:
Beautiful, thank you!

PR #221, opened 2022-08-02
Pass DataPath to install script for upgrades

jshufro:
Uses the flag added in https://github.com/rocket-pool/smartnode-install/pull/74

I believe the logic reordering is safe- the LoadConfig call doesn't seem to have side-effects, and I don't _think_ it will cause errors when run on a fresh install, but I haven't tested it.

### DIFF
--- a/shared/utils/rp/config.go
+++ b/shared/utils/rp/config.go
@@ -20,14 +20,14 @@ func LoadConfigFromFile(path string) (*config.RocketPoolConfig, error) {
 	_, err := os.Stat(path)
 	if os.IsNotExist(err) {
 		return nil, nil
-	} else {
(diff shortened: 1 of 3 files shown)


shared/utils/rp/config.go
```diff
@@ -20,14 +20,14 @@ func LoadConfigFromFile(path string) (*config.RocketPoolConfig, error) {
...
-		}
+	}
 
-		return cfg, nil
+	cfg, err := config.LoadFromFile(path)
+	if err != nil {
+		return nil, err
 	}
+
+	return cfg, nil
```
jshufro, line 30: NFC, just irked me a bit.

jclapis:
That ought to do it, I'll run it through the tests.

PR #222, opened 2022-08-02
764 - Adding option to sign an arbitrary message.

0xfornax:
Adding option to sign an arbitrary message using the node's private key. (EIP-191 personal_sign)
Prints the result in a format that can be directly used on beaconcha.in, but could also be used to sign messages for other use cases.

### DIFF
--- a/rocketpool-cli/node/sign-message.go
+++ b/rocketpool-cli/node/sign-message.go
@@ -43,7 +43,7 @@ func signMessage(c *cli.Context) error {
 	fmt.Printf("Message: %s\n", message)
 	fmt.Printf("Signed data: %s\n\n", response.SignedData)
 
-	if cliutils.Confirm("Do you want to use this message on beaconcha.in?") {
+	if cliutils.Confirm("Do you want to use this message on beaconcha.in? (Prints the message in the expected format)") {
 		fmt.Printf(`{ 
     "address": "%s",
     "msg": "%s",
--- a/rocketpool-cli/node/sign-message.go
+++ b/rocketpool-cli/node/sign-message.go
@@ -52,17 +52,18 @@ func signMessage(c *cli.Context) error {
 	}
 
 	// Print the signature
-	fmt.Printf("Message: %s\n", message)
-	fmt.Printf("Signed data: %s\n\n", response.SignedData)
-
-	if cliutils.Confirm("Do you want to use this message on beaconcha.in? (Prints the message in the expected format)") {
-		bytes, err := json.MarshalIndent(PersonalSignature{status.AccountAddress, message, response.SignedData, fmt.Sprint(signatureVersion)}, "", "  ")
-		if err != nil {
-			return err
-		}
-
-		fmt.Println(string(bytes))
+	formattedSignature := PersonalSignature{
+		Address:   status.AccountAddress,
+		Message:   message,
+		Signature: response.SignedData,
+		Version:   fmt.Sprint(signatureVersion),
 	}
+	bytes, err := json.MarshalIndent(formattedSignature, "", "    ")
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Signed Message:\n\n%s\n", string(bytes))
 
 	return nil
 
--- a/rocketpool-cli/node/sign-message.go
+++ b/rocketpool-cli/node/sign-message.go
@@ -9,7 +9,7 @@ import (
 	cliutils "github.com/rocket-pool/smartnode/shared/utils/cli"
 )
 
-func signMessage(c *cli.Context, message string) error {
+func signMessage(c *cli.Context) error {
 
 	// Get RP client
 	rp, err := rocketpool.NewClientFromCtx(c)
@@ -29,6 +29,7 @@ func signMessage(c *cli.Context, message string) error {
 		return nil
 	}
 
+	message := c.String("message")
 	if message == "" {
 		message = cliutils.Prompt("Please enter the message you want to sign: (EIP-191 personal_sign)", "^.+$", "Please enter the message you want to sign: (EIP-191 personal_sign)")
 	}
@@ -38,12 +39,18 @@ func signMessage(c *cli.Context, message string) error {
 		return err
 	}
 
-	fmt.Printf(`{ 
+	// Print the signature
+	fmt.Printf("Message: %s\n", message)
+	fmt.Printf("Signed data: %s\n\n", response.SignedData)
+
+	if cliutils.Confirm("Do you want to use this message on beaconcha.in?") {
+		fmt.Printf(`{ 
     "address": "%s",
     "msg": "%s",
     "sig": "%s",
     "version": "2"
 }`, status.AccountAddress, message, response.SignedData)
+	}
 
 	return nil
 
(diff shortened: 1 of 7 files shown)


rocketpool-cli/node/sign-message.go
```diff
@@ -0,0 +1,57 @@
...
+		return err
+	}
+
+	if !status.WalletInitialized {
+		fmt.Println("The node wallet is not initialized.")
+		return nil
+	}
+
+	message := c.String("message")
+	if message == "" {
```
jshufro, line 33:
```suggestion
	for message == "" {
```

This would ensure that `message != ""` (ie do minimal validation on stdin)

rocketpool-cli/node/sign-message.go
```diff
@@ -0,0 +1,57 @@
...
+	fmt.Printf("Message: %s\n", message)
+	fmt.Printf("Signed data: %s\n\n", response.SignedData)
+
+	if cliutils.Confirm("Do you want to use this message on beaconcha.in? (Prints the message in the expected format)") {
+		fmt.Printf(`{ 
+    "address": "%s",
+    "msg": "%s",
+    "sig": "%s",
+    "version": "2"
+}`, status.AccountAddress, message, response.SignedData)
```
jshufro, line 52: make a struct and use decorators?

shared/services/wallet/wallet.go
```diff
@@ -291,6 +292,27 @@ func (w *Wallet) Sign(serializedTx []byte) ([]byte, error) {
...
+	fullMessage := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(message), message)
+	hash := crypto.Keccak256Hash([]byte(fullMessage))
+
+	signedMessage, err := crypto.Sign(hash.Bytes(), privateKey)
+	if err != nil {
+		return nil, fmt.Errorf("Error signing message: %w", err)
+	}
+
+	// fix the ECDSA 'v' (see https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v/38909#38909)
+	signedMessage[64] += 27
```
jshufro, line 312: lol, wtf

jshufro (changes requested):

rocketpool-cli/node/sign-message.go
0xfornax: Ok. But this validation was already done by the Prompt regex. (^.+$).

shared/services/wallet/wallet.go
0xfornax: Changed the link to point to a better reference. 
Don't blame me, this += 27 comes from the Yellow paper
image

rocketpool-cli/node/sign-message.go
0xfornax: Not sure I see the benefits. Could you elaborate if you think that's really important?
Btw, I'm changing the version to 1, meaning it's our first signature "spec". The 2 came from what's used by MyCripto and Beaconcha.in. But it can be anything.

rocketpool-cli/node/sign-message.go
jshufro: Isn't `-m` optional?

rocketpool-cli/node/sign-message.go
jshufro: Sure. There's 2 advantages, really:

1. Indentation doesn't break, which is admittedly a nit
2. sprintf is much more error prone and less future proof. This is valid JSON... it's easy to look at it and see that. But if it ever gets more complex, we may miss a stray comma or something. golang's native serializer won't.

jshufro: Basically, something like this:

https://go.dev/play/p/vqE5ALXgVvI

0xfornax: Let me argue a bit:
- Printing as you suggest breaks the identation, making it a single line. I personally prefer 
this (current)
{
   "address":"0x17...",
   "msg":"aasdçflkajdfçlakdsjfçalskdfdkjçalsdkfjaçldfkjaçldfkjaçdlfkajdçflkajdfçlakdfjçlsdkfjaçdslkfjaçdlfkjaçdflkajsdçflkajdsf",
   "sig":"0xd1b...",
    "version":"1"
}

over this, specially with big signatures
{"address":"0x17...","msg":"aasdçflkajdfçlakdsjfçalskdfdkjçalsdkfjaçldfkjaçldfkjaçdlfkajdçflkajdfçlakdfjçlsdkfjaçdslkfjaçdlfkjaçdflkajsdçflkajdsf","sig":"0xd1b1...","version":"1"}
- We have an api response struct that I thought about using for this purpose but we have some extra fields, like "error" that don't need to be printed. Would be weird to create another struct and transfer from one to the other just for printing purposes;
- The version on the struct would be of type uint, but beaconcha.in expects a string, so more steps to prepare the struct for printing;
- It's a very simple json and I don't see it growing complex.

jshufro: You can pretty-print the json with MarshalIndent: https://go.dev/play/p/YdU8c7uBb2c

I agree that printing extra fields feels weird, but I disagree that there's such a thing as a struct too small to bother with. As for not wanting to transfer fields... you already are transferring them into the variadric args of Printf. Transferring them to a new struct is pretty much the same.

Make Version a string if that's what beaconcha.in expects, of course.

0xfornax: Agreed. There you go Patches.

shared/services/wallet/wallet.go
```diff
@@ -291,6 +292,27 @@ func (w *Wallet) Sign(serializedTx []byte) ([]byte, error) {
...
+	fullMessage := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(message), message)
+	hash := crypto.Keccak256Hash([]byte(fullMessage))
+
+	signedMessage, err := crypto.Sign(hash.Bytes(), privateKey)
+	if err != nil {
+		return nil, fmt.Errorf("Error signing message: %w", err)
+	}
+
+	// fix the ECDSA 'v' (see https://medium.com/mycrypto/the-magic-of-digital-signatures-on-ethereum-98fe184dc9c7#:~:text=The%20version%20number,2%E2%80%9D%20was%20introduced)
+	signedMessage[64] += 27
```
jshufro, line 312: based on my read of the above:
```markdown
In most implementations, [the ](https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739)[v](https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739)[ is just 0 or 1 internally](https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739), but 27 was added as arbitrary number for signing Bitcoin messages and Ethereum adapted that as well.
```
Shouldn't this be:
`signedMessage[64] = 27`? To guard against changes in the underlying implementation (keyword above being "most" implementations.

jshufro (approved):

shared/services/wallet/wallet.go
0xfornax: I would stick to the spec and keep adding 27 to this bit. I think using a fixed value makes it incompatible with ethers.js signatures for instance.

0xfornax:
After Patches bravely tried this on mainnet we have proof it works.
image

jclapis:
Tested and it works - it'll need some documentation for things like Beaconcha.in but this is good to go!

PR #225, opened 2022-08-08
Prefer external IPv4 discovery

poupas:
In `#support`, we're seeing node operators with two distinct, but related issues:

  * Bad peering with Nimbus/Besu. This is because on dual-stack (IPv4 and IPv6) nodes, the external IP detection code sometimes finds the IPv4, other times the IPv6. Using IPv6 is not ideal, and will cause peering / network issues.
  * Mysterious container restarts. For example, sometimes a user will update the docker image tag for Grafana or a consensus client. When starting the rocketpool service (e.g. `rocketpool s s`) the `rocketpool_eth1` container will unexpectedly be recreated. We've managed to track this down to the same issue: the external IP detection code switching between IPv4 and IPv6. Every time it changes, the containers will be recreated and restarted.

A potential solution is, when discovering the external IP address, to try IPv4 first. This should improve peering and promote a more consistent view of what the external IP is.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/fatih/color v1.13.0
 	github.com/ferranbt/fastssz v0.0.0-20220103083642-bc5fefefa28b
 	github.com/gdamore/tcell/v2 v2.4.1-0.20210905002822-f057f0a857a1
-	github.com/glendc/go-external-ip v0.0.0-20200601212049-c872357d968e
+	github.com/glendc/go-external-ip v0.1.0
 	github.com/google/uuid v1.3.0
 	github.com/hashicorp/go-version v1.4.0
 	github.com/herumi/bls-eth-go-binary v0.0.0-20211108015406-b5186ba08dc7 // indirect
--- a/shared/services/rocketpool/client.go
+++ b/shared/services/rocketpool/client.go
@@ -65,8 +65,6 @@ const (
 // * Improve peer discovery and node performance
 // * Avoid unnecessary container restarts caused by switching between IPv4 and IPv6
 func getExternalIP() (net.IP, error) {
-	var err error
-
 	// Try IPv4 first
 	ip4Consensus := externalip.DefaultConsensus(nil, nil)
 	ip4Consensus.UseIPProtocol(4)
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/fatih/color v1.13.0
 	github.com/ferranbt/fastssz v0.0.0-20220103083642-bc5fefefa28b
 	github.com/gdamore/tcell/v2 v2.4.1-0.20210905002822-f057f0a857a1
-	github.com/glendc/go-external-ip v0.0.0-20200601212049-c872357d968e
+	github.com/glendc/go-external-ip v0.1.0
```
msf, line 23: should explicitly mention the dependency upgrade change in the commits + PR description

shared/services/rocketpool/client.go
```diff
@@ -60,6 +61,23 @@ const (
 	DebugColor = color.FgYellow
 )
 
+// Get the external IP address. Try finding an IPv4 address first to:
+// * Improve peer discovery and node performance
+// * Avoid unnecessary container restarts caused by switching between IPv4 and IPv6
+func getExternalIP() (net.IP, error) {
```
msf, line 67: I agree that external IP should be "stable" and have a deterministic ordering of obtaining an IP.

however, in this day and age with so much NAT, preferring IPv4 instead of IPv6 doesn't feel like the best option.
For example, I have dual-ip stack and I'd much rather discover that besu has poor ipv6 support and report bugs to besu than using IPv4 ..

I dunno if I can prove factually than ipv4 first is better or the opposite though..

go.mod
jclapis: Concur, I usually don't upgrade libraries without a good reason to avoid supply chain attacks. I'll look through this to see if it provides anything new or introduces anything scary.

shared/services/rocketpool/client.go
jclapis: I know this has bitten some community members several times in the past, so at least for now, preferring IPv4 first will help more people than hurt. The client teams are certainly aware of their IPv6 limitations, so I'm inclined to do this for now in the name of supporting them until they fix it.

go.mod
jclapis: In this case, it looks like the protocol specification flag was introduced in v0.1.0, so this update is required. The diff doesn't show anything alarming: https://github.com/glendc/go-external-ip/compare/c872357d968e..v0.1.0

jclapis:
Alright let's do it and put it into the v1.5.1 test build. Thanks Poupas!

dietzam:
Thank you very much poupas!

PR #228, opened 2022-08-19
Linting pass for overly nested control flow

jshufro:
Also cleans out some sprintf with no-token strs


jclapis:
This took longer to look over than I care to admit - thanks for cleaning it all up.

PR #231, opened 2022-08-26
New command to remove validator keys safely

### LINKED ISSUE #212: Improper key management leads to unintended slashing (Validator 391311)
A Rocket Pool node operator was recently slashed after attempting to wipe keys but failing to clear the keys properly. While it is regarded as user error, Rocket Pool ought to learn from these user errors and prevent similar events in the future.

[Beaconcha.in link to slashed validator #391311](https://beaconcha.in/validator/391311)

[Discord dicussion link - Rocket Pool discord, #Trading channel](https://discord.com/channels/405159462932971535/405163713063288832/990325863235063838)

[Screenshot of discord discussion](https://i.ibb.co/cxnsRsN/Screenshot-from-2022-06-25-16-03-36.png)

Here are two suggestions that could have prevented this loss:

* Ensure the integrity of all 4 keystores. Maybe maintain a lockfile so that only one client (of any type) is running.
* Implement a wallet purge feature like `rocketpool wallet purge` that fully removes any trace of user wallets, this removes the need for the user to attempt manual removal.

0xfornax:
Adding a new command `rocketpool wallet purge-keys` that removes validator keys from all clients.

Closes #212


rocketpool-cli/wallet/purge-keys.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Get & check wallet status
+	status, err := rp.WalletStatus()
+	if err != nil {
+		return err
+	}
+	if !status.WalletInitialized {
+		fmt.Println("The node wallet is not initialized.")
+		return nil
+	}
```
jshufro, line 29: I don't think this should be an early return-
If I initialize a wallet, then spin up a minipool, then run `rm .rocketpool/data/wallet`, then run `rocketpool wallet purge-keys`, my POLA senses would be triggered if smartnode decided not to delete the validator keys just because I manually deleted the node wallet:

```golang
// Check if the wallet has been initialized
func (w *Wallet) IsInitialized() bool {
        return (w.ws != nil && w.seed != nil && w.mk != nil)
}
```

rocketpool-cli/wallet/purge-keys.go
```diff
@@ -0,0 +1,44 @@
...
+		_, err := rp.PurgeKeys()
+		if err != nil {
+			return err
+		}
+
+		fmt.Println("Deleted all validator keys.")
+
+	}
+
+	return nil
```
jshufro, line 42:
```suggestion
	if !cliutils.Confirm("WARNING:\n This command will delete all files related to validator keys. Do you want to continue?") {
		fmt.Println("No action was taken.")
		return nil
	}

	// Rebuild wallet
	_, err := rp.PurgeKeys()
	if err != nil {
		return err
	}

	fmt.Println("Deleted all validator keys.")
	return nil
```

I like the very last scope of the function to be the one that is the "normal" case. I do also think there's some value in informing the user that nothing happened after they said "no"

rocketpool/api/wallet/purge-keys.go
```diff
@@ -0,0 +1,97 @@
...
+				customKeyPath := filepath.Join(customKeyDir, file.Name())
+				err = os.RemoveAll(customKeyPath)
+				if err != nil {
+					return nil, fmt.Errorf("error removing file %s: %w", file.Name(), err)
+				}
+
+			}
+		}
+	}
+	return &response, nil
```
jshufro, line 96:
```suggestion
	if os.IsNotExist(err) || !info.IsDir() {
		// There are no custom keys, so exit early
		return &response, nil
	}

	// Get the custom keystore files
	files, err := ioutil.ReadDir(customKeyDir)
	if err != nil {
		return nil, fmt.Errorf("error enumerating custom keystores: %w", err)
	}

	if len(files) == 0 {
		return &response, nil
	}

	// Process every custom key found
	for _, file := range files {
		// Read the file
		bytes, err := ioutil.ReadFile(filepath.Join(customKeyDir, file.Name()))
		if err != nil {
			return nil, fmt.Errorf("error reading custom keystore %s: %w", file.Name(), err)
		}

		// Deserialize it
		keystore := api.ValidatorKeystore{}
		err = json.Unmarshal(bytes, &keystore)
		if err != nil {
			return nil, fmt.Errorf("error deserializing custom keystore %s: %w", file.Name(), err)
		}

		// Check if it's one of the pubkeys for the minipool
		_, exists := pubkeyMap[keystore.Pubkey.Hex()]
		if !exists {
			// This pubkey isn't for any of this node's minipools so ignore it
			continue
		}
		customKeyPath := filepath.Join(customKeyDir, file.Name())
		err = os.RemoveAll(customKeyPath)
		if err != nil {
			return nil, fmt.Errorf("error removing file %s: %w", file.Name(), err)
		}
	}

	return &response, nil
```
Applying demorgan's law lets us strip 1 layer of indentation, and exiting early when len(files) == 0 lets us strip another, which improves readability a little.

shared/services/rocketpool/wallet.go
```diff
@@ -164,6 +164,22 @@ func (c *Client) RebuildWallet() (api.RebuildWalletResponse, error) {
 	return response, nil
 }
 
+// Rebuild wallet
```
jshufro, line 167: copy pasted from the `rebuild` command?

rocketpool-cli/wallet/purge-keys.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+
+	"github.com/urfave/cli"
+
+	"github.com/rocket-pool/smartnode/shared/services/rocketpool"
+	cliutils "github.com/rocket-pool/smartnode/shared/utils/cli"
+)
+
+func purgeKeys(c *cli.Context) error {
```
jshufro, line 12: Deleting the files isn't enough- the keys are held in memory by the running process. The `validator` container should be restarted on the success of `purgeKeys` (eth2 should be restarted if nimbus is selected).

shared/services/rocketpool/wallet.go
0xfornax: Fixed.  :heh:

rocketpool-cli/wallet/purge-keys.go
0xfornax: You're right. This command should delete validator keys under any condition. I'm removing this check.

rocketpool-cli/wallet/purge-keys.go
0xfornax: Very good call Patches. I'm worried about choosing a certain behavior for each client (ex. restart eth2 only for nimbus) as we would need to track any changes to clients that could have an impact on what needs to be restarted. I'm thinking we should play it safe and restart the whole service. What do you think @jclapis?

0xfornax: Update: Now I'm calling the RestartValidator function which checks what needs to be restarted according to the client being used.

jclapis:
I turned this into a generic "purge" command that also deletes the node wallet and password file so there's no way to accidentally recover the validator keys. If the intent is to make a machine clean to use as a fallback or purge it during a migration, this is a safer thing to do. Thanks Fornax!

PR #232, opened 2022-08-31
Added queue position information to "minipool status" command

RomiRand:
Currently you need to check 3rd party sites to get a minipools position in the activation queue. This could be a bit annoying since it can take weeks.
So I propose offering this information to the NO in the cli.
Based on rocket-pool/rocketpool-go#9


jclapis:
I removed the ETA printout since it is wildly variable based on deposit size, but the queue information is quite useful I think. Thank you for putting this together!

PR #247, opened 2022-09-21
763 - Add ENS support

0xfornax:


0xfornax:
@jclapis this PR has the following changes:
- Smartnode calls that receive addresses (send, set-voting-delegate, set-withdrawal-address) can now receive an ENS name that will be resolved to the corresponding address.;
- A few places where the command result shows addresses (`rp node status` at first) will now show the ENS name resulting from a ReverseResolve call.
- Added a new command `rp wallet set-ens-name`: will set the ENS reverse record name field for the node account to a desired name. (Note: I'm experimenting with using the same function to submit the tx or just estimate the gas cost. Please let me know if you think this approach works here).

jclapis:
Added a few little tweaks, the rest looks good. I'll play with it and get some docs ready for when we release it. Thank you for the hard work on doing this!

PR #251, opened 2022-09-24
Migrate PoW terms to PoS.

0xfornax:
tx mined -> tx included in a block
miner -> validator

### DIFF
--- a/rocketpool/api/node/commands.go
+++ b/rocketpool/api/node/commands.go
@@ -276,7 +276,7 @@ func RegisterSubcommands(command *cli.Command, name string, aliases []string) {
 			{
 				Name:      "wait-and-swap-rpl",
 				Aliases:   []string{"p2"},
-				Usage:     "Swap old RPL for new RPL, waiting for the approval TX hash to be mined first",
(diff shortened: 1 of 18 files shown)


rocketpool/api/node/commands.go
```diff
@@ -276,7 +276,7 @@ func RegisterSubcommands(command *cli.Command, name string, aliases []string) {
 			{
 				Name:      "wait-and-swap-rpl",
 				Aliases:   []string{"p2"},
-				Usage:     "Swap old RPL for new RPL, waiting for the approval TX hash to be mined first",
+				Usage:     "Swap old RPL for new RPL, waiting for the approval TX hash to be include in a block first",
```
0xobjectObject, line 279: typo, pls fix

0xobjectObject (changes requested):

rocketpool/api/node/commands.go
0xfornax: Nice catch, fixed.

jclapis:
Thanks Fornax!

PR #256, opened 2022-10-03
Fix comment

jshufro:
High value diff, I know.


jclapis:
There's a joke about GitPOAP in here somewhere

PR #257, opened 2022-10-03
Updated Nethermind, see PR in smartnode-install

0xobjectObject:


jclapis:
I was about to do a client pass, thanks for this!

PR #258, opened 2022-10-09
Update oDAO fee distributor calculation to accurately reflect rEth-owned amounts

jshufro:


jclapis:
Math checks out, thanks for the fix Patches.

PR #260, opened 2022-10-14
Make daemon sending deposit transaction optional

xrchz:
If the transaction is not sent, its RLP encoding is printed out instead.
Depends on https://github.com/rocket-pool/rocketpool-go/pull/10.


jclapis:
LGTM, hope this works for you!

PR #261, opened 2022-10-16
Use LondonSigner for api node sign

xrchz:


jclapis:
I love one-liner improvements, well done. Thanks!

PR #263, opened 2022-10-20
Remove total order assumptions from recover/rebuild key search

jshufro:


jclapis:
Fixed in an earlier commit, thanks for bringing this to my attention!

PR #266, opened 2022-11-01
Adapt snapshot struct to receive a response of type "Any"

### LINKED ISSUE #265: Error requesting DAO proposals
```sh
$ rocketpool -v

rocketpool version 1.6.5

$ rocketpool --debug network dao-proposals

To API:
docker exec rocketpool_api /go/bin/rocketpool   --maxFee 0.000000 --maxPrioFee 0.000000 --gasLimit 0   api service get-client-status
API Out:
{"status":"success","error":"","ecManagerStatus":{"primaryEcStatus":{"isWorking":true,"isSynced":true,"syncProgress":1,"error":""},"fallbackEnabled":false,"fallbackEcStatus":{"isWorking":false,"isSynced":false,"syncProgress":0,"error":""}},"bcManagerStatus":{"primaryEcStatus":{"isWorking":true,"isSynced":true,"syncProgress":1,"error":""},"fallbackEnabled":false,"fallbackEcStatus":{"isWorking":false,"isSynced":false,"syncProgress":0,"error":""}}}

Your Smartnode is currently using the Ethereum Mainnet.

To API:
docker exec rocketpool_api /go/bin/rocketpool --ignore-sync-check  --maxFee 0.000000 --maxPrioFee 0.000000 --gasLimit 0   api network dao-proposals
API Out:
{"status":"error","error":"could not decode snapshot response: json: cannot unmarshal object into Go struct field SnapshotProposalVote.data.votes.choice of type int","accountAddress":"0x0000000000000000000000000000000000000000","votingDelegate":"0x0000000000000000000000000000000000000000","activeSnapshotProposals":null,"proposalVotes":null}

error after requesting dao proposals: could not decode snapshot response: json: cannot unmarshal object into Go struct field SnapshotProposalVote.data.votes.choice of type int
```

0xfornax:
Snapshot Choice has a type "Any", so it needs to be treated as an interface{} in Go.

To parse and show results, so far I've seen that interface can be either float64, []interface{} (which is actually []float64), or a map[string] interface{} (which becomes map[string] float64). All these cases should work correctly now.

Fix #265

### DIFF
--- a/shared/types/api/node.go
+++ b/shared/types/api/node.go
@@ -421,7 +421,7 @@ type SnapshotResponse struct {
 	}
 }
 type SnapshotProposalVote struct {
-	Choice   int            `json:"choice"`
+	Choice   interface{}    `json:"choice"`
 	Voter    common.Address `json:"voter"`
 	Proposal struct {
 		Id string `json:"id"`
--- a/rocketpool-cli/network/dao-proposals.go
+++ b/rocketpool-cli/network/dao-proposals.go
@@ -100,6 +100,9 @@ func getActiveDAOProposals(c *cli.Context) error {
 					case map[string]interface{}:
 						choice := proposalVote.Choice.(map[string]interface{})
 						for index, weight := range choice {
+							if votedChoices != "" {
+								votedChoices += ", "
+							}
 							votedChoices += fmt.Sprintf("%s: %.2f", proposal.Choices[int(choice[index].(float64))], weight)
 						}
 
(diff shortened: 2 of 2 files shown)


shared/types/api/node.go
```diff
@@ -421,7 +421,7 @@ type SnapshotResponse struct {
 	}
 }
 type SnapshotProposalVote struct {
-	Choice   int            `json:"choice"`
+	Choice   interface{}    `json:"choice"`
```
jshufro, line 424: I think this can/should be json.RawMessage

jclapis:
Thanks for taking this one Fornax!

PR #274, opened 2022-12-08
Update pipe wait logic around calling to Docker

krobertson:
This PR updates some of the instrumentation around streaming stdout/stderr from docker-compose to the user. Previously, this code was updated to use cmd.Start rather than cmd.Run, which was a correct change, but it also instrumented a WaitGroup around the goroutines handling the IO, but how it was done was not the way it was intended.

Previously, `pipeToStdOut` and `pipeToStdErr`, would themselves call to `wg.Add`, however it would be calling to Add from within the new goroutine. In Go, a new goroutine does not immediately begin executing, so it would typically end up that `wg.Wait()` would be reached before either of the new goroutines would call to `wg.Add`. It is typically recommended to call `wg.Add` before any goroutines are started to ensure they're accounted for.

To illustrate, see the quick Go example. The majority of the time, you will never see `IN` printed. However, if you move `wg.Add(1)` to be called before the gorotuine, it will then work as intended.

https://go.dev/play/p/UpurAEafzNo

Second, I removed the use of the `WaitGroup` because it already was not being used, and for the most part, is not necessary. The primary way the pipes would be closed is when the processes closes, so it is really only necessary to call to `cmd.Wait`. With the wait group, it would be doing a double wait. Wait for the IO, then when thats done, it'd get to `cmd.Wait` and the process is already done, so it'd just move on.

Additionally, I moved the call to setup the goroutines for streaming the IO to before Start is called, just to ensure its all setup when the command is ran.

### DIFF
--- a/shared/services/rocketpool/client.go
+++ b/shared/services/rocketpool/client.go
@@ -14,7 +14,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/a8m/envsubst"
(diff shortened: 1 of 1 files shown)


jclapis:
Alright, thank you for taking a look at this! Let me do some quick tests on it just to verify, then I'll merge it in.

jclapis:
Looks good - thank you for fixing this!

PR #276, opened 2022-12-10
mention that ens names are supported in send command's help

smile5131:


jclapis:
Sure, thanks for the update!

PR #277, opened 2022-12-13
Add --force to the export-eth1-data cmd.

### LINKED ISSUE #230: Incremental exports fail due to df checks
When a NO backs up their eth1 data to an external device, they should be able to do incremental backups due to the underlying utility being rsync.

However, smartnode does a `df` check on the device and errors out if there isn't enough free space:
```
Chain data size: 612 GiB Target dir free space: 321 GiB
Your target directory does not have enough space to hold the chain data. Please free up more space and try again.
```

We should add a flag to skip the `df` check

0xfornax:
Adds a --force flag to the export-eth1-data command to ignore the free space check on the target folder.

Closes #230


jclapis:
Easy enough, ship it. Thanks Fornax.

PR #280, opened 2022-12-29
Removed auto respond to challenges

knoshua:
This turns challenges into an activity check that allows removal of zombie nodes that are still technically online but have been abandoned and are potentially running outdated software versions that are no longer able to contribute to consensus.

### DIFF
--- a/rocketpool/watchtower/respond-challenges.go
+++ b/rocketpool/watchtower/respond-challenges.go
@@ -88,45 +88,7 @@ func (t *respondChallenges) run() error {
 	}
 
 	// Log
-	t.log.Printlnf("Node %s has an active challenge against it, responding...", nodeAccount.Address.Hex())
+	t.log.Printlnf("Node %s has an active challenge against it.", nodeAccount.Address.Hex())
 
-	// Get transactor
-	opts, err := t.w.GetNodeAccountTransactor()
-	if err != nil {
-		return err
-	}
-
-	// Get the gas limit
-	gasInfo, err := trustednode.EstimateDecideChallengeGas(t.rp, nodeAccount.Address, opts)
-	if err != nil {
-		return fmt.Errorf("Could not estimate the gas required to respond to the challenge: %w", err)
-	}
-
-	// Print the gas info
-	maxFee := eth.GweiToWei(WatchtowerMaxFee)
-	if !api.PrintAndCheckGasInfo(gasInfo, false, 0, t.log, maxFee, 0) {
-		return nil
-	}
-
-	// Set the gas settings
-	opts.GasFeeCap = maxFee
-	opts.GasTipCap = eth.GweiToWei(WatchtowerMaxPriorityFee)
-	opts.GasLimit = gasInfo.SafeGasLimit
-
-	// Respond to challenge
-	hash, err := trustednode.DecideChallenge(t.rp, nodeAccount.Address, opts)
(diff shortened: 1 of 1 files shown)


darcius:
Hey @knoshua. Removing this auto-respond would then require every single oDAO member to manually check their logs every 7 days to see if they have been challenged even if their node is operating perfectly fine and reporting correctly.

It also opens up a pretty serious attack vector where an attacker with some ETH could wait until a period of the year like Xmas/New Years (or big events like Ethdenver), when it is possible good oDAO node operators might be away for more than 7 days and execute challenges across all of the oDAO in the hopes they'd get a few of them away. With a challenge being 1 ETH and no auto-responder happening, it would be trivial for a serious competitor to attempt this even with the oDAO in complete consensus and operating normally.

The oDAO 'member kick' function currently exists for the purpose of removing zombie oDAO node operators who refuse to update to the latest consensus smart node software or they cannot be reached, but their node is still running. They can also have their bond slashed completely via this method.

knoshua:
Thank you for your comment @darcius .

I believe it is very important that oDAO members monitor their node at all the times, even during holidays or big events. Not paying attention for 7 days straight is an issue in and of itself in my opinion. 

I acknowledge that the possibility of kicking members exists, but this is only available to the oDAO and not the community. I also feel like that the system in it's current state is very all or nothing and lacks incentives for oDAO nodes to perform their duties. If you can find the time, I would appreciate your thoughts on some community developed ideas to improve this over on the governance forum: https://dao.rocketpool.net/t/improving-odao-participation/1333

Valdorff:
@darcius: I think @knoshua is on to a real issue here. We saw with Etherscan that the oDAO did _not_ use the member kick function. The challenge function was used and a system that was limping along (outdated, not able to make trees, etc) would have auto-responded just fine.

I 100% agree with you about 2 issues however:
- There would need to be obvious notifications (rooting around in logs daily is silly)
- There would need to be a defense against multiple challenges

For the first, I think the technical work required to catch the challenge event and send an email or similar is pretty trivial. We can do that.

For the latter, it would require a small smart contract change, but we could make challenges scale by the number of open challenges. For example, first challenge is 1 ETH. A challenge while a challenge is open is 10 ETH. A challenge while 2 challenges are open is 100 ETH. Etc. This makes using challenges as an attack wildly expensive and ineffective. Would you be open to including something like that in the release after Atlas, and then (once that's in place) putting in a change to make responding non-automatic?

darcius:
Thanks for the input guys. I do agree that having a zombie node that still responds to challenges is a possibility, but the kick function would be more than sufficient to handle this and it would provide additional oversight + the opportunity to reach out to the oDAO member before their bond is burned is a necessity.

With Etherscans situation, everything was working as intended. There was no need to use the kick function immediately as we had reached out to them and was trying to assist them in getting their node in order, this oversight is important, we can't always assume the worst about members immediately. The challenge was made by the community and that was great, it was working as intended since there node was not responding and it would have seen them removed.

We could look at adding some additional measures post Atlas, but I genuinely believe the current system is fine as is. It enables good oDAO members to run their setups without micromanaging weekly checks that could easily be used as an attack vector. Should they fail to do so and their node isn't up, they can be challenged. If their node is up, responding but not up to date and submitting bad trees, they can be kicked by the other members after we have reached out to them which provides an important additional level of oversight.

Issue #249, opened 2022-09-23
Timezone detection on installation fails to recognize correct timezone


superphiz:
IIRC Timezone detection worked fine for a long time, but in the two most recent installations I've done it offers 'Etc/UTC' as the detected timezone while the correct timezone has an offset from UTC.

`rocketpool node register`

`Would you like to detect your timezone automatically? [y/n]`
`y`

`The detected timezone is 'Etc/UTC', would you like to register using this timezone? [y/n]`

jclapis:
Timezone detection uses your IP's geolocation - if it can't reach the geolocation server or that server doesn't have your IP listed, it will show up as Etc/UTC as a catch-all.

superphiz:
I'm offering a 250 USDC bounty to anyone who completes this issue with two solutions:
1. Better locale determination, mine is never picked up, it always shows "Etc/UTC". Ubuntu installers hit it perfectly every time and I like that.
2.  a listing of acceptable timezones that the user can choose from. The first dialog should show the continent, then the locales within that continent.

0xHodja:
#282 

Get request seems to fail on the current "https://freegeoip.app/json/" due to missing an API key.

Proposing solution to change from "https://freegeoip.app/json/" to a working API "https://ipinfo.io/json" which is public without requiring an API key.

Although this does not fulfil the second criteria, it should fix the majority of first-pass time zone location string issues.

superphiz:
I'll pay 1/2 the bounty for this part of the job as soon as it gets merged. Thanks Hodja!

sambacha:
> I'm offering a 250 USDC bounty to anyone who completes this issue with two solutions:
> 
> 1. Better locale determination, mine is never picked up, it always shows "Etc/UTC". Ubuntu installers hit it perfectly every time and I like that.
> 2. a listing of acceptable timezones that the user can choose from. The first dialog should show the continent, then the locales within that continent.

Configuring your locale can and does effect more than TZ, it can effect sorting, math, etc. 

Nodes should always be configured to UTC, since by definition UNIX epoch time is UTC.

Additionally if you are wanting to configure your locale and ignore the above, you must also configure your NTP pool connection information to whatever is the largest pool near your locale 


Also TZ can leak and reduce privacy

0xHodja:
> Configuring your locale can and does effect more than TZ, it can effect sorting, math, etc.
> 
> Nodes should always be configured to UTC, since by definition UNIX epoch time is UTC.
> 
[...]

PR #282, opened 2022-12-31
Changed API to a working one for IP geolocation

0xHodja:
#249 

Get request seems to fail on the current "https://freegeoip.app/json/" due to missing an API key.

Proposing solution to change from "https://freegeoip.app/json/" to a working API "https://ipinfo.io/json" which is public without requiring an API key.

Although this does not fulfil the second criteria, it should fix the majority of first-pass time zone location string issues.


jclapis:
LGTM :+1:

PR #283, opened 2022-12-31
Changing the IP Geolocation API and list timezone options

### LINKED ISSUE #249: Timezone detection on installation fails to recognize correct timezone

0xfornax:
- Changing the API to a free one with more accurate results;
- If the user wants to manually inform, first ask for continent/country/city to show a filtered list with options.

Closes #249


jclapis:
I adjusted this so that it prints the actual country and region lists the user can select from - that way they aren't just selecting blindly. An interaction looks like this:

```
Checking your system's timezone...
The detected timezone is 'Etc/UTC', would you like to register using this timezone? [y/n]
n

You will now be prompted to enter a timezone.
For a complete list of valid entries, please use one of the "TZ database name" entries listed here: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

List of valid countries / continents:
Africa
America
Antarctica
Arctic
Asia
Atlantic
Australia
Brazil
Canada
Chile
Etc
Europe
Indian
Mexico
Pacific
US

Please enter a country / continent from the list above:
US

List of valid regions:
Alaska
Aleutian
Arizona
Central
East-Indiana
Eastern
Hawaii
Indiana-Starke
Michigan
Mountain
Pacific
Samoa

Please enter a region from the list above:
Eastern

Using timezone US/Eastern.
```

PR #287, opened 2023-01-03
Remove %s and add examples.

### LINKED ISSUE #281: Service Config > Execution Client fails to complete string
I'm currently syncing geth but planning to switch to an externally managed client. When I go to the Externally Managed > Websocket URL dialog, I find that my current client is identified as "%s"

While you're on this screen, it's ALWAYS helpful to see example inputs, like "http://192.168.1.100:8545" or whatever, it just makes interpreting the fields easier.
image

And oh my god, I just pasted an image right from my clipboard onto this screen and it picked it up. very cool.

0xfornax:
Fix string with dangling %s, and adding connection examples as requested on issue 281.
Closes #281.


jclapis:
:+1:

PR #288, opened 2023-01-03
Adding missing containers to be restarted.

### LINKED ISSUE #284: shared/services/config/external-configs.go - external client URL changes for eth2 does not trigger node/api/watchtower restarts
https://github.com/rocket-pool/smartnode/blob/master/shared/services/config/external-configs.go 

Taking LH as an example:
```golang
		HttpUrl: config.Parameter{
			ID:                   "httpUrl",
			Name:                 "HTTP URL",
			Description:          "The URL of the HTTP Beacon API endpoint for your external client.\nNOTE: If you are running it on the same machine as the Smartnode, addresses like `localhost` and `127.0.0.1` will not work due to Docker limitations. Enter your machine's LAN IP address instead.",
			Type:                 config.ParameterType_String,
			Default:              map[config.Network]interface{}{config.Network_All: ""},
			AffectsContainers:    []config.ContainerID{config.ContainerID_Eth1},
			EnvironmentVariables: []string{"CC_API_ENDPOINT"},
			CanBeBlank:           false,
			OverwriteOnUpgrade:   false,
		},
```

note that `AffectsContainers` only points to eth1 (which, itself, I'm not sure why)... however, updating this directly impacts, at the very least, rocketpool_api, and possibly also watchtower/node, since they check the BN URL when querying sync status.

The impact of this is that updating an external BN URL causes the eth1 container to restart but the api/node/watchtower services continue to query the old URL. Since the VC has an env variable, it is at least recreated, though it is also absent from the list when exiting the tui:

image

image

While we're talking about it, I think `NewExternalExecutionConfig` also wants to restart watchtower

0xfornax:
Closes #284 .


jclapis:
Hmm, I'm gonna have to do this for Nimbus and Lodestar in v1.8 too... I should do that real quick before merging more PRs.

PR #289, opened 2023-01-04
Reduce the number of Snapshot API calls

0xfornax:
Currently, we're making calls to the Snapshot API every 5 mins. 
- Adding a timeout for these connections;
- Using a cached value, updated every 6 hours.


jclapis:
Sounds like a good addition, anything to lower the node's processing time!

PR #291, opened 2023-01-05
Fix the nextAccount value after recovering keys.

### LINKED ISSUE #290: Bug after recovering wallet and validator keys - next_account value should be latest index + 1
I noticed a small bug after recovering a wallet and validator keys.

After the validator keys are recovered, the wallet file is currently storing the value "next_account" using the latest index of the validators, meaning that when there is one rebuilt validator (index = 0 ) the next_account is set as 0 when it should be 1.

Steps to reproduce:
- Remove wallet and validator files.
- Use `rocketpool wallet recover` to recover one validator
- Inspect the file `.rocketpool/data/wallet` and see that the next_account does not reflect the desired value of 1 but instead is still at 0.

0xfornax:
Fix the next_account value after recovering keys (should be walletIndex + 1).

Closes #290

### DIFF
--- a/shared/services/wallet/validator.go
+++ b/shared/services/wallet/validator.go
@@ -211,7 +211,7 @@ func (w *Wallet) SaveValidatorKey(key ValidatorKey) error {
 
 	// Update account index
 	if key.WalletIndex > w.ws.NextAccount {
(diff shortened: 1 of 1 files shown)


jclapis:
Good catch, thank you Fornax!

PR #296, opened 2023-01-19
Ignore sync check when signing messages.

### LINKED ISSUE #295: Allow signing messages without synced clients
The rescue node will require a signed message from the node wallet to identify the user requesting access has a Rocket Pool node.

Currently, the signMessage function is checking the client sync state. This is a problem as users of the rescue node would not always have synced clients.

0xfornax:
The rescue node will use a signed message to identify the user requesting access has a Rocket Pool node and will require a signed message from the node wallet.

Currently, the signMessage function is checking the client sync state. This is a problem as users of the rescue node would not always have synced clients.

Closes #295

### DIFF
--- a/rocketpool/api/node/sign-message.go
+++ b/rocketpool/api/node/sign-message.go
@@ -13,10 +13,6 @@ import (
 )
 
 func signMessage(c *cli.Context, message string) (*api.NodeSignResponse, error) {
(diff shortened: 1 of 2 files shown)


jshufro:
lgtm, thanks for the fast turnaround

PR #303, opened 2023-01-26
Fix panic when the word from the wordlist being compared is short

### LINKED ISSUE #302: Wallet test-recovery panic: runtime error: slice bounds out of range [:4] with length 3
Steps to reproduce the error:

$ rocketpool w t
Enter 24
Enter 'gunn'
panic: runtime error: slice bounds out of range [:4] with length 3
  
image

0xfornax:
Closes #302


jclapis:
Ah man I forgot about this one, thanks for fixing it!

PR #304, opened 2023-01-29
Use W3s upstream

SN9NV:
This dependency has been updated, so it's not necessary to maintain our own fork, and override.

I also updated the go.mod go version, 1.13 technically is wrong and would not work because some code needs at least 1.18. I saw no reason not to specify 1.19. Then I ran `go mod tidy` to update to the new schema.

I only tested that the `rocketpool-cli` could be built and it runs at least on Linux x86-64. I'm not really sure how to test anything else.


jclapis:
Ok, glad to see they finally updated it. I'll have to run this through the dependency tests because we updated a bunch of stuff in v1.9 for Atlas.

PR #305, opened 2023-02-05
Fix terminal resize

SN9NV:
Fixes a bug where if you start the config wizard with a small terminal, and you get the message that the terminal is too small, and then you maximize the terminal, it won't redraw, so you have to Ctrl+C and run the configuration again.

Using `BeforeDraw`, fixes this problem because it's deciding to render the "too small" message or wizard when the terminal has been resized, so you don't need to trigger another render to get this to update.


jclapis:
Seems reasonable, good find!

PR #306, opened 2023-02-05
Fix subprocess writing to stdout/err

SN9NV:
Hopefully this is the last time we do this lol.

The problem with `io.Copy` is that when the `cmd` is not given an `io.Writer` to write to, it will create it's own, and it will close this when it exits, and the `Read` function does not like this.

So in this case, the proper thing to do is to give it an `io.Writer` to write to, and we can use `os.Stdout/err` for this. When we set the writers, the `cmd` does not not close them on exit, which is what we want in the case for `os.Stdout/err`.

### DIFF
--- a/shared/services/rocketpool/client.go
+++ b/shared/services/rocketpool/client.go
@@ -4,9 +4,7 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	"io"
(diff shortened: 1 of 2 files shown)


jclapis:
Fingers crossed...

Issue #268, opened 2022-11-05
Feature Request: Integrated smartnode client updating


superphiz:
Currrently, users must access some resource to find the latest smartnode client version, while this is readily available in the docs, but it does require accessing an external resource to update the smartnode client. The ideal operation would be a client that is able to download its own updates from the cli without accessing resources in a different location.

For example, `rocketpool update` ought to check the current version against the latest available version, and if a newer version is available, it ought to replace the existing binary with the newer version and assist the user through the update process.

PR #311, opened 2023-02-12
Add support for CLI update command

aliask:
As requested by Phiz in #268, I have added support for a `rocketpool update cli` command which will:
- check what the latest version is via the Github API
- compare with local version, and download it if it's newer
- verify the PGP signature of the downloaded binary
- replace existing client if all the above goes well

There are some assumptions regarding the filenames in the release but these appear to be reasonably widely referenced already.

Optional flags included for this command:
- `--force` will attempt to update, even if Github version is same/older
- `--skip-signature-verification` will prevent any checks against the PGP signature
- `--yes` will automatically confirm the update

### DIFF
--- a/rocketpool-cli/update/update.go
+++ b/rocketpool-cli/update/update.go
@@ -1,0 +1,311 @@
+package update
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/blang/semver/v4"
+	"github.com/urfave/cli"
+	"golang.org/x/crypto/openpgp"
+
+	"github.com/rocket-pool/smartnode/shared"
+	"github.com/rocket-pool/smartnode/shared/services/config"
+	cfgtypes "github.com/rocket-pool/smartnode/shared/types/config"
+	cliutils "github.com/rocket-pool/smartnode/shared/utils/cli"
+)
+
+// Settings
+const (
+	ExporterContainerSuffix         string = "_exporter"
+	ValidatorContainerSuffix        string = "_validator"
+	BeaconContainerSuffix           string = "_eth2"
+	ExecutionContainerSuffix        string = "_eth1"
+	NodeContainerSuffix             string = "_node"
+	ApiContainerSuffix              string = "_api"
+	WatchtowerContainerSuffix       string = "_watchtower"
+	PruneProvisionerContainerSuffix string = "_prune_provisioner"
+	EcMigratorContainerSuffix       string = "_ec_migrator"
+	clientDataVolumeName            string = "/ethclient"
+	dataFolderVolumeName            string = "/.rocketpool/data"
+
+	PruneFreeSpaceRequired uint64 = 50 * 1024 * 1024 * 1024
+	dockerImageRegex       string = ".*/(?P<image>.*):.*"
+	colorReset             string = "\033[0m"
+	colorBold              string = "\033[1m"
+	colorRed               string = "\033[31m"
+	colorYellow            string = "\033[33m"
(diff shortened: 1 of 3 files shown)


rocketpool-cli/update/update.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	for _, v := range entity.Identities {
+		fmt.Printf("Signature verified. Signed by: %s\n", v.Name)
+	}
+	return nil
+}
+
+// Update the Rocket Pool CLI
+func updateCLI(c *cli.Context) error {
```
SN9NV, line 70: I see a few main sections to this function. I think it would be easier to see the main steps if they are split into their own functions. It would also help add context to the errors. I see the same error message repeated/similar to another one. What do you think?

SN9NV (commented):
This is an awesome feature! I hope it gets merged soon.

rocketpool-cli/update/update.go
aliask: Yes, I think that's a fair comment. I've split out the initial version check, and the download/verify sections into their own functions, and I think `updateCLI` is pretty readable now :)

SN9NV:
Seems great. I tested the various flags and it seems to be working as expected.

There's just these two issues from `golangci-lint`:
```
update/update.go:135:14: Error return value of `output.Seek` is not checked (errcheck)
                output.Seek(0, io.SeekStart)
                           ^
update/update.go:16:2: SA1019: "golang.org/x/crypto/openpgp" is deprecated: this package is unmaintained except for security fixes. New applications should consider a more focused, modern alternative to OpenPGP for their specific task. If you are required to interoperate with OpenPGP systems and need a maintained package, consider a community fork. See https://golang.org/issue/44226. (staticcheck)
        "golang.org/x/crypto/openpgp"
        ^
```

Not critical things, but I think probably good to have a look at.

rocketpool-cli/update/update.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Do the switcheroo - move `rocketpool-vX.X.X` to the location of the current Rocketpool Client
+	err = os.Remove(ex)
+	if err != nil {
+		return fmt.Errorf("error while removing old rocketpool binary: %w", err)
+	}
+	err = os.Rename(fileName, ex)
+	if err != nil {
+		return fmt.Errorf("error while writing new rocketpool binary: %w", err)
+	}
```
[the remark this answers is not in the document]
SN9NV: When replacing the binary, the key is the the inode is different to the original executable. Since we have a new file created during the download process. So this is pretty safe. If we were using something else like truncate/copy, then we might still have the original inode, which can be problematic for the OS to deal with.

aliask:
I've added a check for the `Seek()`, but the openpgp deprecation is a tricky one.
I couldn't find any third party libraries offering a drop-in functionality, but I'm not sure it's really that critical:
- Despite being _officially_ deprecated, it appears to still be being updated - the last release was only a couple of weeks ago
- Other major projects are still depending on it, including [geth](https://github.com/ethereum/go-ethereum/blob/master/.golangci.yml#L53)

SN9NV:
> I've added a check for the `Seek()`, but the openpgp deprecation is a tricky one. I couldn't find any third party libraries offering a drop-in functionality, but I'm not sure it's really that critical:
> 
>     * Despite being _officially_ deprecated, it appears to still be being updated - the last release was only a couple of weeks ago
> 
>     * Other major projects are still depending on it, including [geth](https://github.com/ethereum/go-ethereum/blob/master/.golangci.yml#L53)

Looks good. Yeah I think it will be fine and we can replace the library if there's any CVEs. I'm pretty sure it will be perfectly fine.

SN9NV:
I rate it's good to go. :rocket:

PR #318, opened 2023-02-25
chore: update url for checkpoint sync docs

dabdab-wrt:
Seems like the url in the docs have been updated to:
https://docs.rocketpool.net/guides/node/config-docker.html#beacon-chain-checkpoint-syncing

Not sure if there's already a duplicate commit, please feel free to close if so.


jclapis:
Thank you!

PR #324, opened 2023-03-21
Fix deprecated ioutil file methods

acontry:
Applied the following changes:
* `ioutil.WriteFile` -> `os.WriteFile` ([docs](https://pkg.go.dev/io/ioutil#WriteFile))
* `ioutil.ReadFile` -> `os.ReadFile` ([docs](https://pkg.go.dev/io/ioutil#ReadFile))
* `ioutil.ReadAll` -> `io.ReadAll` ([docs](https://pkg.go.dev/io/ioutil#ReadAll))

These are all no-op changes since the deprecated methods were calling the underlying `os` and `io` methods.

Fixing deprecated `ioutil.ReadDir` is slightly more involved, so did not address that one in this PR.

### DIFF
--- a/rocketpool-cli/node/utils.go
+++ b/rocketpool-cli/node/utils.go
@@ -3,7 +3,7 @@ package node
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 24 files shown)


jclapis:
LGTM, thanks!

PR #331, opened 2023-03-23
Migrate deprecated `ioutil.ReadDir` method calls

acontry:
Migrate all `ioutil.ReadDir` -> `os.ReadDir` as recommended per the deprecation note. Note that `os.ReadDir` returns a different struct than the deprecated method. Both structs support `Name()` and `IsDir()` methods, so this should be a drop-in replacement based on a read of the code.


jclapis:
Thank you!

PR #342, opened 2023-04-09
Change "Expose API Port" from bool to a ternary choice.

### LINKED ISSUE #334: "Expose RPC/API Ports" should allow finer grain control
I'd change these to dropdown boxes with ternary options
1) Off
2) Locally
3) Externally

For 2 a `ports:` mapping of 
```yaml
ports:
  - 127.0.0.1:${RPC_PORT}:${RPC_PORT}
```

will prevent other devices on the network from being able to establish a connection to the port.

For 3 the `ports:` mapping we have today will allow all devices on the network to access the port, ie
```yaml
ports:
  - ${RPC_PORT}:${RPC_PORT}
```

This is especially crucial for reverse-hybrid mode operators on a VPS. Currently they have to either run their validator in the docker network, or do other fancy networking stuff to make the BN api port accessible to the validator, but not the world.

Leaving it open to the world leaves the operator vulnerable to tips/mev theft.

0xfornax:
Closes #334 

Hey @jclapis and @jshufro! Considering I'm changing the type for the "Export API port" parameter from bool to a choice between "closed", "localhost" or "external".  The external option exposes the port to the user's local network or to the internet if they're on a VPS.

I'm creating a migration from 1.9.1 with this setup:
false => closed
true => open to localhost

Users will need to manually select the "external" option if they want to.

### DIFF
--- a/shared/services/config/consensus-common-config.go
+++ b/shared/services/config/consensus-common-config.go
@@ -58,7 +58,7 @@ func NewConsensusCommonConfig(cfg *RocketPoolConfig) *ConsensusCommonConfig {
 		Value:       config.RPC_OpenLocalhost,
 	}, {
 		Name:        "Open to External hosts",
-		Description: "Allow connections from external hosts. This is safe if you're running your node on your local network. If you're a VPS user this would expose your node to the internet and could make your it vulnerable to MEV/tips theft",
+		Description: "Allow connections from external hosts. This is safe if you're running your node on your local network. If you're a VPS user, this would expose your node to the internet and could make it vulnerable to MEV/tips theft",
 		Value:       config.RPC_OpenExternal,
 	}}
 
@@ -119,7 +119,7 @@ func NewConsensusCommonConfig(cfg *RocketPoolConfig) *ConsensusCommonConfig {
 		OpenApiPort: config.Parameter{
 			ID:                   OpenApiPortID,
 			Name:                 "Expose API Port",
-			Description:          "Enable this to expose your Consensus client's API port to your localhost or to your local network, so other machines can access it too.",
+			Description:          "Select an option to expose your Consensus client's API port to your localhost or to your local network, so other machines can access it too.",
 			Type:                 config.ParameterType_Choice,
 			Default:              map[config.Network]interface{}{config.Network_All: defaultOpenBnApiPort},
 			AffectsContainers:    []config.ContainerID{config.ContainerID_Eth2},
--- a/shared/services/config/consensus-common-config.go
+++ b/shared/services/config/consensus-common-config.go
@@ -17,7 +17,7 @@ const defaultGraffiti string = ""
 const defaultCheckpointSyncProvider string = ""
 const defaultP2pPort uint16 = 9001
 const defaultBnApiPort uint16 = 5052
-const defaultOpenBnApiPort string = "closed"
+const defaultOpenBnApiPort string = string(config.RPC_Closed)
 const defaultDoppelgangerDetection bool = true
 
 // Env var names
--- a/shared/services/config/consensus-common-config.go
+++ b/shared/services/config/consensus-common-config.go
@@ -107,7 +107,7 @@ func NewConsensusCommonConfig(cfg *RocketPoolConfig) *ConsensusCommonConfig {
 		OpenApiPort: config.Parameter{
 			ID:                   OpenApiPortID,
 			Name:                 "Expose API Port",
-			Description:          "Select an option to expose your Consensus client's API port to your localhost or to your local network, so other machines can access it too.",
+			Description:          "Select an option to expose your Consensus client's API port to your localhost or external hosts on the network, so other machines can access it too.",
 			Type:                 config.ParameterType_Choice,
 			Default:              map[config.Network]interface{}{config.Network_All: defaultOpenBnApiPort},
 			AffectsContainers:    []config.ContainerID{config.ContainerID_Eth2},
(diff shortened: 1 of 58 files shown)


shared/services/config/mev-boost-config.go
```diff
@@ -100,6 +100,20 @@ func NewMevBoostConfig(cfg *RocketPoolConfig) *MevBoostConfig {
...
+	}, {
+		Name:        "Open to Localhost",
+		Description: "Allow connections from this host only",
+		Value:       config.RPC_OpenLocalhost,
+	}, {
+		Name:        "Open to External hosts",
+		Description: "Allow connections from external hosts. This is safe if you're running your node on your local network. If you're a VPS user, this would expose your node to the internet",
+		Value:       config.RPC_OpenExternal,
+	}}
+
```
jshufro, line 116: let's put this slice in a single place. It gets used quite a lot, might as well be DRY

jshufro: I wouldn't be opposed to breaking types/config/types.go out into types/config/ports.go or something, and sticking it there as
```golang
// RPCModes returns each supported RPCMode as a config.ParameterOption in a slice.
// If passed, warningOverride is used as the Description field of the config.ParameterOption for the 'external' setting.
func PortModes(warningOverride string) []config.ParameterOption {
...
}
```

Just make sure to return a new slice each time because otherwise you have to deal with pass-by-reference crap for warningOverride.

shared/services/config/migration/v191-manager.go
```diff
@@ -0,0 +1,48 @@
...
+		return fmt.Errorf("expected a executionCommon setting named `%s` but it didn't exist", keyOpenPorts)
+	}
+
+	// Update the config
+	if openRPCPorts == "true" {
+		configSection[keyOpenPorts] = config.RPC_OpenLocalhost.String()
+	} else {
+		configSection[keyOpenPorts] = config.RPC_Closed.String()
+	}
+	serializedConfig[configKeyString] = configSection
```
jshufro, line 46:
```suggestion
```
https://go.dev/play/p/w4odHcyxWRd

Concise reason this isn't necessary: a map is actually a pointer to a map, a map of a map is a pointer to map of pointers to maps of pointers. configSection is a pointer to a map of pointers, and that pointer isn't mutated when you modify it.

jshufro (changes requested):

shared/services/config/mev-boost-config.go
0xfornax: Done.

jshufro, line 116: Can this also be a function call, or is it special?

0xfornax: Nothing special, just forgot about this one.

shared/types/config/port_modes.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if rpcMode == RPC_OpenLocalhost {
+		return fmt.Sprintf("127.0.0.1:%s", ports)
+	}
+
+	return ""
+}
+
+func PortModes(warningOverride string) []ParameterOption {
```
jshufro, line 37: Maybe not something that would pass joe's sniff test, but if you make this variadic then the default path looks cleaner.

ie
```golang
func PortModes(warningOverride ...string) []ParameterOption {
    var warning string
    if len(warningOverride) == 0 {
        warning = "Allow connections from external hosts. This is safe if you're running your node on your local network. If you're a VPS user, this would expose your node to the internet"
    } else {
        warning = warningOverride[0]
    }
```

then, elsewhere, instead of
`conf.PortModes("")` you just call `conf.PortModes()` unless there is an override, in which case you pass it as before.

jshufro (changes requested):
I'm not 100% certain the compile-time string(RPCMode_Closed) thing works, but it _should_.

Other than the `git mv` and relocation of `type RPCMode string`, suggestions are "optional". Use your own discretion or ask joe what he prefers.

shared/services/config/consensus-common-config.go
```diff
@@ -17,7 +17,7 @@ const defaultGraffiti string = ""
 const defaultCheckpointSyncProvider string = ""
 const defaultP2pPort uint16 = 9001
 const defaultBnApiPort uint16 = 5052
-const defaultOpenBnApiPort bool = false
+const defaultOpenBnApiPort string = "closed"
```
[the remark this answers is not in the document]
0xfornax: Nice catch. Changing to string(config.RPC_Closed). There were a few extra places too.

jshufro (approved):

shared/types/config/port_modes.go
jclapis: I did a little cleanup to make it more palatable but didn't make it variadic cause comparing against "" has parity with other places in the code... wen refactor?

jclapis:
So I made a few updates, but have a comment before I merge it:

I changed the migration stuff so an old `true` maps to `RPC_OpenExternal` instead of `RPC_OpenLocalhost`. The reason is that people who currently use `true` as it exists today enjoy being able to access those ports on different machine in the subnet, and making it `RPC_OpenLocalhost` would be a breaking change. I'd much rather introduce the feature and suggest everyone on a VPS switch to `RPC_OpenLocalhost` instead, as this is an intentional change in behavior.

jshufro:
Changes look good @jclapis but I disagree with using `OpenExternal` as the migration path for `true`. I think it selects for the worst case scenario- a VPS user who has his tips/mev stolen.

The breaking change will surely cause confusion but people are more likely to go read the patch notes when something breaks, rather than when something creates a security risk but doesn't break.

jclapis:
Yeah, after talking to the guys about this it sounds like we'll go with the breaking change route but that means putting this into v1.10 instead of v1.9.4 (which may or may not end up existing anyway).

PR #350, opened 2023-04-18
Trimming spaces from form fields.

### LINKED ISSUE #349: Error caused by an extra space at the end the fallback URL string
Waqwaq had the following error when trying to use his fallback:

Could not get client status: error connecting to fallback EC at [xxxx ]: parse "xxxx   ": invalid port ":8545   " after host

That was caused by an extra space at the end of the string.
image

The smartnode should trim the string to avoid such errors in the future.

0xfornax:
Trims white spaces from the beginning and end of form fields and on the wizard.

Closes #349

### DIFF
--- a/rocketpool-cli/service/config/config-form.go
+++ b/rocketpool-cli/service/config/config-form.go
@@ -3,6 +3,7 @@ package config
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


jclapis:
I'm sure this was fun to find, thanks Fornax.

PR #353, opened 2023-04-23
Change the rp-minipool-performance files so they only report on duties while a minipool was opted in to the smoothing pool

jshufro:
Prior to this commit, a node that opts into the smoothing pool in the last epoch of the 6300 epoch interval will still have its performance for all 6300 epochs recorded in the performance file.

This is misleading:

Imagine a hypothetical node with a single minipool that is not in the SP

At interval epoch 0, it goes offline
At interval epoch 6200, it comes back online and joins the smoothing pool
At interval end, the current rewards file will show this minipool as having missed 6200 attestations, with a participation rate of 0.015873016. However, it made 100/100 attestations while in the smoothing pool.

This change will make the performance file show that the total attestations (failed + succeeded) is 100, and the participation rate is 1, which is more accurate than showing 6200 missed attestations and 100 successful attestations and a participation rate of 0.015.

NB: This cannot be merged haphazardly, and may actually need to be implemented on v6 instead of v5.

### DIFF
--- a/shared/services/rewards/generator-impl-v5.go
+++ b/shared/services/rewards/generator-impl-v5.go
@@ -840,47 +840,54 @@ func (r *treeGeneratorImpl_v5) checkDutiesForSlot(attestations []beacon.Attestat
 
 		// Get the RP committees for this attestation's slot and index
 		slotInfo, exists := r.intervalDutiesInfo.Slots[attestation.SlotIndex]
-		if exists {
-			rpCommittee, exists := slotInfo.Committees[attestation.CommitteeIndex]
-			if exists {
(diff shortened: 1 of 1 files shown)


shared/services/rewards/generator-impl-v5.go
```diff
@@ -840,47 +840,54 @@ func (r *treeGeneratorImpl_v5) checkDutiesForSlot(attestations []beacon.Attestat
...
+
+			// Check if this minipool was opted into the SP for this block
+			nodeDetails := r.nodeDetails[validator.NodeIndex]
+			if blockTime.Sub(nodeDetails.OptInTime) < 0 || nodeDetails.OptOutTime.Sub(blockTime) < 0 {
+				// Not opted in
+				continue
+			}
+
+			// Mark this duty as completed
+			validator.CompletedAttestations[attestation.SlotIndex] = true
```
jshufro, line 877: this line moved down, from above the optin/optout checks. 

the rest of this diff blob is whitespace, because the nesting was making my head hurt

PR #361, opened 2023-05-08
BN client Committees response optimizations

jshufro:
* Change []beacon.Committee to an interface to reduce copying on the hot path
* Allow buffered decoding in the BN std-http-client
* Reuse []string buffers for bn committee responses
* Reuse json.Decoder objects (which have internal buffers) across calls to the beacon node's committees endpoint

### DIFF
--- a/shared/services/beacon/client.go
+++ b/shared/services/beacon/client.go
@@ -65,10 +65,23 @@ type BeaconBlock struct {
 	ExecutionBlockNumber uint64
 }
 
-type Committee struct {
-	Index      uint64
-	Slot       uint64
-	Validators []string
+// Committees is an interface as an optimization- since committees responses
+// are quite large, there's a decent cpu/memory improvement to removing the
+// translation to an intermediate storage class.
+//
+// Instead, the interface provides the access pattern that smartnode (or more
+// specifically, tree-gen) wants, and the underlying format is just the format
+// of the Beacon Node response.
+type Committees interface {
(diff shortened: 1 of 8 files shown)


shared/services/beacon/client.go
```diff
@@ -65,10 +65,23 @@ type BeaconBlock struct {
...
+//
+// Instead, the interface provides the access pattern that smartnode (or more
+// specifically, tree-gen) wants, and the underlying format is just the format
+// of the Beacon Node response.
+type Committees interface {
+	// Index returns the index of the committee at the provided offset
+	Index(int) uint64
+	// Slot returns the slot of the committee at the provided offset
+	Slot(int) uint64
+	// Validators returns the list of validatorsof the committee at the
```
jshufro, line 80:
```suggestion
	// Validators returns the list of validators of the committee at the
```

shared/services/beacon/client.go
```diff
@@ -65,10 +65,23 @@ type BeaconBlock struct {
 	ExecutionBlockNumber uint64
 }
 
-type Committee struct {
-	Index      uint64
-	Slot       uint64
-	Validators []string
+// Committees in an interface as an optimization- since committees responses
```
jshufro, line 68:
```suggestion
// Committees is an interface as an optimization- since committees responses
```

jclapis:
Alrighty, in it goes!

PR #364, opened 2023-05-16
[NEW] Add offline mode

brad-u410:
This enables operators to still run the API and Node containers, making it easier to keep the node operator key entirely offline.

I'm currently testing this for internal use, but wanted to see if this was worth getting merged upstream.

### DIFF
--- a/shared/services/config/rocket-pool-config.go
+++ b/shared/services/config/rocket-pool-config.go
@@ -57,6 +57,9 @@ type RocketPoolConfig struct {
 
 	IsNativeMode bool `yaml:"-"`
 
+	Offline     config.Parameter `yaml:"offline"`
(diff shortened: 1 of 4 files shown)


jclapis:
This is actually going to be useful for an upcoming refactor we have planned. Offline is kind-of a misnomer if what it really means is "simulate all of the transactions but don't actually sign and submit anything to the network"; if I were to merge this upstream I'd probably replace the name with something like "simulation mode".

brad-u410:
@jclapis - in my case, the node operator key _is_ kept offline. When it needs to sign transactions, those are transferred to the airgapped machine, signed, exported and broadcast.

However, that caused two issues:
1. Most the of CLI no longer works, even for informational commands
2. I don't know what transactions I _would_ have been automatically signing via the smartnode container, as I can't run that service

PR #371, opened 2023-06-06
[FIX] replace broken link in error message

paul-leydier:
Hi everyone!

There's an error log when the Smartnode fails to delete a non-existing folder when changing the network in the config:
> The Smartnode could not automatically change networks for you, so you will have to run the steps manually. Please follow the steps laid out in the Node Operator's guide (https://docs.rocketpool.net/guides/node/mainnet.html).

Unless I'm mistaken, in this message, the link is broken.
I suggest replacing it with the [Creating a Standard Rocket Pool Node with Docker
](https://docs.rocketpool.net/guides/node/docker.html) page, but feel free to update my change with another page if it could make more sense.

### DIFF
--- a/rocketpool-cli/service/service.go
+++ b/rocketpool-cli/service/service.go
@@ -331,7 +331,7 @@ func configureService(c *cli.Context) error {
 
 			err = changeNetworks(c, rp, fmt.Sprintf("%s%s", prefix, ApiContainerSuffix))
 			if err != nil {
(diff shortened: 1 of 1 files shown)


jclapis:
This is a good point - it used to refer to this page, which has since been moved:

https://docs.rocketpool.net/guides/legacy/upgrading.html

I think we probably need a *real* page for safely swapping networks that this can point to; not just one for upgrading from Rocket Pool betas. I'll look into adding one and update this link. Thanks for the report!

Issue #380, opened 2023-07-19
smartnode installer shouldn't pipe install.sh into a shell


jshufro:
`curl | sh`
or `wget | sh`

are both antipatterns, as an SSL error will cause a silent failure (treated by smartnode as a success)

instead, use golang's http library to download the install.sh script, and if it succeeds, hand the result off to sh using the command's stdin fd

PR #381, opened 2023-07-19
change the way smartnode downloads and runs the install script to catch errors with the download better

jshufro:
closes https://github.com/rocket-pool/smartnode/issues/380


jclapis:
Truly some ancient stuff, glad to see it get improved.

PR #395, opened 2023-10-04
Replace eth1/eth2 with execution/consensus

harpocryptes:
eth1/eth2 is an outdated terminology. There's been some work before to rename, e.g. https://github.com/rocket-pool/smartnode/commit/727d7ea2a5987a0f99896422de06a9b36c82e794 but there are some eth1/eth2 left, so the current state is inconsistent. This attempts to fix that.


charlesweiss528 (changes requested):
I disagree. I do not approve these transactions. I find them to be fraudulent. I'm flagging these transactions. I'm Flagging,Stopping and Reversing these transactions immediately.

PR #410, opened 2023-11-21
Stake eth on behalf

0xfornax:
Adding a `withdraw-eth` cli command to allow NOs to withdraw unused ETH that was previously staked on behalf of the node.
`node deposit` also checks for that ETH balance when creating a new minipool.
`node status` also displays that information.


rocketpool-cli/pdao/commands.go
```diff
@@ -1059,7 +1059,7 @@ func RegisterCommands(app *cli.App, name string, aliases []string) {
 									{
 										Name:      "submit-balances-frequency",
 										Aliases:   []string{"sbf"},
-										Usage:     fmt.Sprintf("Propose updating the %s setting; %s", protocol.SubmitBalancesFrequencySettingPath, blockCountUsage),
+										Usage:     fmt.Sprintf("Propose updating the %s setting; %s", protocol.SubmitBalancesFrequencySettingPath, uintUsage),
```
jclapis, line 1062: We probably want this (and `SubmitPricesFrequencySettingPath` below) to be durations instead of uints; since the actual value is going to be in seconds, duration makes more sense here.

rocketpool-cli/pdao/get-settings.go
```diff
@@ -68,9 +68,9 @@ func getSettings(c *cli.Context) error {
 	fmt.Printf("\tNode Penalty Quorum:        %.2f%%\n", response.Network.NodePenaltyThreshold*100)
 	fmt.Printf("\tPenalty Size:               %.2f%%\n", response.Network.PerPenaltyRate*100)
 	fmt.Printf("\tBalance Submission Enabled: %t\n", response.Network.IsSubmitBalancesEnabled)
-	fmt.Printf("\tBalance Submission Freq:    %d Blocks\n", response.Network.SubmitBalancesFrequency)
+	fmt.Printf("\tBalance Submission Freq:    %d Seconds \n", response.Network.SubmitBalancesFrequency)
```
jclapis, line 71: Same comment, if we treat this as a duration we can print it in human-readable format instead of a raw number of seconds

rocketpool/api/node/deposit.go
```diff
@@ -92,13 +99,23 @@ func canNodeDeposit(c *cli.Context, amountWei *big.Int, minNodeFee float64, salt
...
-	wg1.Go(func() error {
-		ethBalanceWei, err := node.GetNodeDepositCredit(rp, nodeAccount.Address, nil)
-		if err == nil {
-			response.CreditBalance = ethBalanceWei
-		}
-		return err
-	})
+	if isHoustonDeployed {
+		wg1.Go(func() error {
+			ethBalanceWei, err := node.GetNodeCreditAndBalance(rp, nodeAccount.Address, nil)
```
jclapis, line 104: I think this probably wants to be `GetNodeUsableCreditAndBalance` for deposit since getting the total amount doesn't help you if you can't use it?

rocketpool/node/collectors/odao-collector.go
```diff
@@ -76,16 +76,13 @@ func (collector *OdaoCollector) Collect(channel chan<- prometheus.Metric) {
 	blockNumberFloat := float64(state.ElBlockNumber)
 	pricesBlockFloat := float64(state.NetworkDetails.PricesBlock)
 	effectiveRplStakeBlockFloat := pricesBlockFloat
-	latestReportableBlockFloat := float64(state.NetworkDetails.LatestReportablePricesBlock)
```
jclapis, line 79: Taking this out instead of adding a `isHoustonDeployed` check on it will remove it from the Smartnode while Atlas is still live, and we probably want to keep it around during that.

shared/services/config/smartnode-config.go
```diff
@@ -572,6 +587,24 @@ func NewSmartnodeConfig(cfg *RocketPoolConfig) *SmartnodeConfig {
 			config.Network_Holesky: {},
 		},
 
+		previousRocketNetworkPricesAddresses: map[config.Network][]common.Address{
+			config.Network_Mainnet: {
+				common.HexToAddress("0x751826b107672360b764327631cC5764515fFC37"),
+			},
+			config.Network_Prater:  {},
+			config.Network_Devnet:  {},
+			config.Network_Holesky: {},
```
jclapis, line 596: We're gonna need to put the Holesky addresses in here and below since eventually that'll get upgraded to Houston and price / balance reporting will stop working with v1.12 on that while Atlas is still live.

shared/types/api/node.go
```diff
@@ -52,6 +52,9 @@ type NodeStatusResponse struct {
 	EthMatchedLimit                          *big.Int        `json:"ethMatchedLimit"`
 	PendingMatchAmount                       *big.Int        `json:"pendingMatchAmount"`
 	CreditBalance                            *big.Int        `json:"creditBalance"`
+	CreditAndBalance                         *big.Int        `json:"creditAndBalance"`
```
jclapis, line 55: Probably want to call this like `CreditAndDonatedBalance` or something since the above is `CreditBalance`, it sounds weird if it's just `CreditAndBalance`... Not gonna die on that hill though. Same to the other fields below.

0xfornax: I was keeping it consistent with names used on the smart contracts. But will change here.

Issue #401, opened 2023-11-02
[bug] ~/.rocketpool/user-settings.yml gets truncated when the drive is full


jshufro:
This has bugged me for a while, and I've always known how to fix it, but today i helped then Nth NO fix it and would like to not do so again.

Instead of writing to ~/.rocketpool/user-settings.yml on exit, `rocketpool service config` should write to a different file, say, `~/.rocketpool/user-settings.yml.tmp` and then use [os.Rename](https://pkg.go.dev/os#Rename) to replace `~/.rocketpool/user-settings.yml`

Unix file semantics tell us that we can trivially assume that if the first write succeeds, then the move will use less space, not more, so we can assume it'll work, and prevent truncation of the target file.

`Even within the same directory, on non-Unix platforms Rename is not an atomic operation.` thankfully, we only support unix!

PR #411, opened 2023-11-22
Prevent truncation of user-settings.yml when disk is full by writing to /tmp first and moving the result to user-settings.yml

jshufro:
closes https://github.com/rocket-pool/smartnode/issues/401


0xfornax (approved):
LGTM 
thanks Patches!

PR #414, opened 2023-11-30
Phase out web3.storage usage

0xfornax:
- Won't fail if web3.storage uploads fail (will continue with submission flow using the CID)
- Adds rescuenode.com as a trusted provider for rewards trees files
- Users can provide one (or more) custom download URLs

### DIFF
--- a/rocketpool/watchtower/submit-rewards-tree-rolling.go
+++ b/rocketpool/watchtower/submit-rewards-tree-rolling.go
@@ -598,8 +598,9 @@ func (t *submitRewardsTree_Rolling) generateTree(rp *rocketpool.RocketPool, stat
 		// Upload the rewards tree file
 		t.printMessage("Uploading to Web3.Storage and submitting results to the contracts...")
 		cid, err := t.uploadFileToWeb3Storage(wrapperBytes, compressedRewardsTreePath, "compressed rewards tree")
+		// Don't return on error as we're phasing out. Just keep the CID for the submission and users will be able to fetch from different providers
 		if err != nil {
-			return fmt.Errorf("Error uploading Merkle tree to Web3.Storage: %w", err)
+			t.printMessage(fmt.Sprintf("Error uploading Merkle tree to Web3.Storage: %w", err))
 		}
 		t.printMessage(fmt.Sprintf("Uploaded Merkle tree with CID %s", cid))
 
--- a/shared/services/config/smartnode-config.go
+++ b/shared/services/config/smartnode-config.go
@@ -30,6 +30,7 @@ const (
 	PrimaryRewardsFileUrl              string = "https://%s.ipfs.dweb.link/%s"
 	SecondaryRewardsFileUrl            string = "https://ipfs.io/ipfs/%s/%s"
 	GithubRewardsFileUrl               string = "https://github.com/rocket-pool/rewards-trees/raw/main/%s/%s"
+	RescueNodeRewardsFileUrl           string = "https://rescuenode.com/rewards-trees/%s/%s"
 	FeeRecipientFilename               string = "rp-fee-recipient.txt"
 	NativeFeeRecipientFilename         string = "rp-fee-recipient-env.txt"
 )
@@ -78,6 +79,9 @@ type SmartnodeConfig struct {
 
 	// Mode for acquiring Merkle rewards trees
 	RewardsTreeMode config.Parameter `yaml:"rewardsTreeMode,omitempty"`
+
+	// Custom URL to download a rewards tree
+	RewardsTreeCustomUrl config.Parameter `yaml:"rewardsTreeCustomUrl,omitempty"`
 
 	// URL for an EC with archive mode, for manual rewards tree generation
 	ArchiveECUrl config.Parameter `yaml:"archiveEcUrl,omitempty"`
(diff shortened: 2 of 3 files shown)


shared/services/rewards/utils.go
```diff
@@ -305,12 +304,23 @@ func DownloadRewardsFile(cfg *config.RocketPoolConfig, interval uint64, cid stri
 				}
 			}
 
+			// Verify if the downloaded file has the expected CID
+			calculatedCid, err := SingleFileDirIPFSCid(writeBytes, rewardsTreePath)
```
jshufro, line 308: `writeBytes` here is always decompressed, but the ipfs cid will be on the .zst-suffixed, compressed file.

We should deserialize the decompressed bytes into a RewardsFIle / MinipoolPerformanceFile struct and check the merkle root instead- standardizing on a specific zstd implementation (that specifically says _not to hash its output for similarity_) is going to be fragile.

jshufro: (
ideally we recompute the root so we are certain that:
1. every proof is correct
2. every proof is present
)

shared/services/rewards/utils.go
```diff
@@ -305,12 +304,39 @@ func DownloadRewardsFile(cfg *config.RocketPoolConfig, interval uint64, cid stri
...
 			}
 
+			deserializedRewardsFile, err := DeserializeRewardsFile(writeBytes)
+			if err != nil {
+				return fmt.Errorf("Error deserializing file %s: %w", rewardsTreePath, err)
+			}
+
+			// Get the original merkle root
+			downloadedRoot := deserializedRewardsFile.GetHeader().MerkleRoot
+
```
jshufro, line 314:
```suggestion
			downloadedRoot := deserializedRewardsFile.GetHeader().MerkleRoot
			deserializedRewardsFile.GetHeader().MerkleRoot = ""

```

Alright hear me out on this one. Obviously, MerkleRoot gets overwritten on L316 when we call `generateMerkleTree()`, so why would we want to erase it first if it's about to get erased anyway?

Two reasons-
1. If the behavior changes, this is defensive
2. It improves the 'readability' of this function, since it makes it more obvious that the original header had a parsed root, we nuke it, and then it gets regenerated.

jshufro:
Thanks, I feel much better about this than requiring the computed cid to match!

shared/services/rewards/utils.go
0xfornax: Sounds good.

0xfornax:
> The watchtower was still referring to the files as `rp-whatever.json.zst` in the CID calculation, but since you removed `zstd` compression that's no longer valid, so I pushed up a commit that gets rid of the compressed filenames. Note this means that if we want to upload these to IPFS, they'll have to be exactly what is sent to the CID creation routine later which means right now it's all uncompressed JSON. The files are going to get fairly big, so I wouldn't recommend a free-tier pinning service for them; especially for Holesky where they're pushed daily.

I was experimenting with removing zstd and also removing proofs from the rewards file as those can be rebuilt. But I agree it's too soon for those changes and this PR should be only about removing web3.storage so I reintroduced the zstd compression.

rocketpool-cli/node/claim-rewards.go
```diff
@@ -75,7 +75,7 @@ func nodeClaimRewards(c *cli.Context) error {
 		// Download the files
 		for _, missingInterval := range missingIntervals {
 			fmt.Printf("Downloading interval %d file... ", missingInterval.Index)
-			err := rprewards.DownloadRewardsFile(cfg, missingInterval.Index, missingInterval.CID, false)
+			err := rprewards.DownloadRewardsFile(cfg, missingInterval.Index, missingInterval.CID, missingInterval.MerkleRoot, false)
```
jshufro, line 78: kind of feels like this should be `missingInterval.DownloadRewardsFile(cfg, false)`

missingInterval is part of `rprewards` anyway.

rocketpool/watchtower/submit-rewards-tree-rolling.go
```diff
@@ -506,23 +504,14 @@ func (t *submitRewardsTree_Rolling) runRewardsIntervalReport(client *rocketpool.
...
-			t.log.Printf("error uploading Merkle tree to Web3.Storage: %s. Calculating CID", err.Error())
-			c, err := rprewards.GetCIDForSerializedFile(fileBytes, filepath.Base(compressedRewardsTreePath))
-			if err != nil {
-				return fmt.Errorf("error getting CID for file %s: %w", compressedRewardsTreePath, err)
-			}
-			cid = c.String()
-			t.printMessage(fmt.Sprintf("Calculated rewards file CID %s", cid))
-		} else {
-			t.log.Printlnf("%s Uploaded Merkle tree with CID %s", t.logPrefix, cid)
+			return fmt.Errorf("error submitting rewards snapshot: %w", err)
```
jshufro, line 509: reword this comment to specify there was an error while calculating the CID.

"submitting rewards snapshot" is vague anyway- doesn't specify whether it's submitted to web3storage vs the odao contracts

edit: elsewhere you use `fmt.Errorf("Error getting CID for file %s: %w", compressedRewardsTreePath, err)`

shared/services/config/external-configs.go
```diff
@@ -460,15 +460,14 @@ func NewExternalTekuConfig(cfg *RocketPoolConfig) *ExternalTekuConfig {
...
 		DoppelgangerDetection: config.Parameter{
-			ID:                   DoppelgangerDetectionID,
-			Name:                 "Enable Doppelgänger Detection",
-			Description:          "If enabled, your client will *intentionally* miss 1 or 2 attestations on startup to check if validator keys are already running elsewhere. If they are, it will disable validation duties for them to prevent you from being slashed.",
-			Type:                 config.ParameterType_Bool,
-			Default:              map[config.Network]interface{}{config.Network_All: defaultDoppelgangerDetection},
-			AffectsContainers:    []config.ContainerID{config.ContainerID_Validator},
-			EnvironmentVariables: []string{"DOPPELGANGER_DETECTION"},
-			CanBeBlank:           false,
-			OverwriteOnUpgrade:   false,
```
jshufro, line 471: is this diff because of the envsubst branch? You should merge master into your branch, or rebase, and this will be dropped.

shared/services/config/smartnode-config.go
```diff
@@ -327,10 +327,10 @@ func NewSmartnodeConfig(cfg *RocketPoolConfig) *SmartnodeConfig {
...
 		},
 
-		Web3StorageApiToken: config.Parameter{
-			ID:                 "web3StorageApiToken",
-			Name:               "Web3.Storage API Token",
-			Description:        "[orange]**For Oracle DAO members only.**\n\n[white]The API token for your https://web3.storage/ account. This is required in order for you to upload Merkle rewards trees to Web3.Storage at each rewards interval.",
+		RewardsTreeCustomUrl: config.Parameter{
+			ID:                 "rewardsTreeCustomUrl",
+			Name:               "Rewards tree custom download URLs",
+			Description:        "[orange]**Only used if you want an extra source to download rewards tree files.**[white]\n\nThe smartnode will automatically try to download rewards tree files from sources like IPFS and GitHub. Use this field if you want to provide a extra URL for the download (multiple URLs can be provided using ';' as separator).\nUsers don't need to trust any of the file sources as the content will need to match what was voted by the oDAO.",
```
jshufro, line 333:
```suggestion
			Description:        "[orange]**Only used if you want an extra source to download rewards tree files.**[white]\n\nThe smartnode will automatically try to download rewards tree files from sources like IPFS and GitHub. Use this field if you want to provide an extra URL for the download (multiple URLs can be provided using ';' as separator).\nUsers don't need to trust any of the file sources as the content will be verified against the on-chain results from the oDAO.",
```

shared/services/rewards/utils.go
```diff
@@ -310,46 +311,50 @@ func DownloadRewardsFile(cfg *config.RocketPoolConfig, interval uint64, cid stri
...
-	}
-	root, err := dagFmtr.Add(file, "", fsMap)
-	if err != nil {
-		return cid.Cid{}, fmt.Errorf("error adding rewards file to DAG: %w", err)
+			// Compare the merkle roots to see if the original is correct
+			if downloadedRoot != calculatedRoot {
+				return fmt.Errorf("the merkle root from the downloaded file %s is not correct", rewardsTreePath)
+			}
+
+			if calculatedRoot != expectedRoot.String() {
```
jshufro, line 336: these are hex, so capitalization is actually kinda arbitrary, and i suggest using `strings.EqualFold` here and on L332

jshufro: Slightly out of scope but it would be nice to delete the `else` on L292 and outdent this whole block.

```golang
if thing {
    foo()
    continue;
} else {
    bar();
}
```
makes it less obvious that bar is unconditional when !thing

prefer:
```golang
if thing {
    foo()
    continue;
}

bar();
```

shared/services/config/external-configs.go
0xfornax: I did merge master into this branch, but needed to adjust for the envsubst removal of the EnvironmentVariables.

PR #416, opened 2023-12-01
Continue submission when the file upload fails

0xfornax:
This PR would be the first stage of the web3.storage phase out. At this point, we would continue to have w3s as a dependency allowing uploads to fail gracefully and continue the submission process by calculating the CID locally. 
A second stage would completely remove the w3s dependency.  #414


jclapis:
This LGTM as a short-order fix, I'd suggest testing it on Holesky with a manually modded `hosts` file to simulate a web3.storage API failure. If it all works as expected then let's put it into the next release.

PR #418, opened 2023-12-11
Dedup calls to LoadConfig

jshufro:
We often call `cliutils.PrintNetwork(rp)`, which in turn calls `rp.LoadConfig()`, which then discards the result before it returns. The same functions often proceed to call `rp.LoadConfig()` again.

This isn't ideal for two reasons:
1) If LoadConfig ever has side-effects, it may not be reentrant
2) We parse stuff twice, which isn't slow but it isn't fast either.

Instead we can just add a function on RocketPoolConfig to get the network, and change `cliutils.PrintNetwork` to take that as an argument.


0xfornax (approved):
LGTM

PR #419, opened 2023-12-11
Print rescue node plugin info even when the EC is syncing

jshufro:
Since many people use Rescue Node to resync eth1 (or change clients), it would be ideal if `rocketpool node status` always printed the Rescue Node details when both the wallet is initialized and the plugin is enabled.

This will conflict with https://github.com/rocket-pool/smartnode/pull/418 so maybe merge that one first and I'll rebase

### DIFF
--- a/addons/rescue_node/addon.go
+++ b/addons/rescue_node/addon.go
@@ -138,7 +138,7 @@ func (r *RescueNode) PrintStatusText(nodeAddr common.Address) {
 	// Check the Username
 	usernameNodeAddr, err := r.getCredentialNodeId()
 	if err != nil {
-		fmt.Printf("%s%w%s", colorRed, err, colorReset)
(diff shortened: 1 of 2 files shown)


0xfornax:
#418 was merged if you want to do that rebase

jshufro:
> #418 was merged if you want to do that rebase

All set :)

0xfornax:
Would be nice to separate errors from the smartnode and the add-on to avoid things like this 
image

jshufro:
> Would be nice to separate errors from the smartnode and the add-on to avoid things like this image

Definitely not ideal, looking into it.

jshufro:
Pushed a fix. I forgot that `fmt.Errorf` wants %w for wrapping, and Printf wants %v for stringifying. Added linebreaks.

govet catches this. We really need to make an effort to get better CI static analysis.

0xfornax:
Thanks for the fixes. lgtm

PR #420, opened 2023-12-11
Remove envsubst

jshufro:
MUST be released hand-in-hand with https://github.com/rocket-pool/smartnode-install/pull/113

Doing prometheus.yml first, so I can get feedback early.


shared/services/rocketpool/client.go
```diff
@@ -300,32 +301,23 @@ func (c *Client) UpdatePrometheusConfiguration(settings map[string]string) error
...
 
-	// Write the actual Prometheus config file
-	err = os.WriteFile(prometheusConfigPath, contents, 0664)
+	// Create the output file
+	runtimeFile, err := os.OpenFile(prometheusConfigPath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0664)
 	if err != nil {
-		return fmt.Errorf("Could not write Prometheus config file to %s: %w", shellescape.Quote(prometheusConfigPath), err)
+		return fmt.Errorf("Could not open Prometheus config file %s for writing: %w", shellescape.Quote(prometheusConfigPath), err)
 	}
-	err = os.Chmod(prometheusConfigPath, 0664)
```
jclapis, line 326: I vaguely remember having to add this to work around `umask`, does the new stuff obviate that?

jshufro: I feel like the only way you'd end up with a file that isn't 0664 would be if one were created by an earlier version of smartnode. Even the old code passed 0664 to WriteFile. Do you remember more details?

jclapis: All I remember is that someone had an unusual `umask` config and it wouldn't actually write the file to `0664` until we explicitly chmod'd it with this line, and we figured no harm no foul. We can leave it out and see if anyone complains again, at least you'll have more info from a direct source then :P

PR #432, opened 2024-01-11
Add local file abstraction for treegen and fix compression bugs with cid calculation

jshufro:
Introducing a new interface allows us to abstract the process of writing/reading tree and performance files from disk, as well as compressing the files and calculating their CIDs.

This resolves at least one bug where CIDs are calculated with incorrect filenames.

Read shared/services/rewards/files.go first. This is the new abstraction.
See it in action in the rest of the files. The rules are:
1. When downloading or generating a tree, we call `NewLocalFile` to create the struct, `Write` when we want to write it, `CompressedCid()` to calculate the cid and write the compressed file to disk.
2. When reading a file from disk we use `ReadLocalRewardsFile` or `ReadLocalMinipoolPerformanceFile` which reads the files and parses the bytes in one go. It assumes the file is _not_ compressed, as the only code that operates over compressed files is the tree downloader, which decompresses ad-hoc. We can feasibly change the contents of the file and overwrite it with `Write()`, but we don't actually do that anywhere.

New tests are passing:
```
$ go test -test.v .
=== RUN   TestFilesFromTree
    files_test.go:11: TestFilesFromTree using tempdir /tmp/TestFilesFromTree4014754575/001
--- PASS: TestFilesFromTree (0.00s)
=== RUN   TestCompressionAndCids
    files_test.go:68: TestCompressionAndCids using tempdir /tmp/TestCompressionAndCids271413775/001
--- PASS: TestCompressionAndCids (0.21s)
PASS
ok      github.com/rocket-pool/smartnode/shared/services/rewards        (cached)
```

### DIFF
--- a/rocketpool/watchtower/generate-rewards-tree.go
+++ b/rocketpool/watchtower/generate-rewards-tree.go
@@ -248,28 +248,24 @@ func (t *generateRewardsTree) generateRewardsTreeImpl(rp *rocketpool.RocketPool,
 	// Create the JSON files
 	rewardsFile.SetMinipoolPerformanceFileCID("---")
 	t.log.Printlnf("%s Saving JSON files...", generationPrefix)
-	minipoolPerformanceBytes, err := rewardsFile.GetMinipoolPerformanceFile().Serialize()
-	if err != nil {
-		t.handleError(fmt.Errorf("%s Error serializing minipool performance file into JSON: %w", generationPrefix, err))
-		return
-	}
-	wrapperBytes, err := rewardsFile.Serialize()
-	if err != nil {
-		t.handleError(fmt.Errorf("%s Error serializing proof wrapper into JSON: %w", generationPrefix, err))
-		return
-	}
+	localMinipoolPerformanceFile := rprewards.NewLocalFile[rprewards.IMinipoolPerformanceFile](
+		rewardsFile.GetMinipoolPerformanceFile(),
+		t.cfg.Smartnode.GetMinipoolPerformancePath(index, true),
+	)
+	localRewardsFile := rprewards.NewLocalFile[rprewards.IRewardsFile](
+		rewardsFile,
+		t.cfg.Smartnode.GetRewardsTreePath(index, true),
+	)
 
 	// Write the files
-	path := t.cfg.Smartnode.GetRewardsTreePath(index, true)
-	minipoolPerformancePath := t.cfg.Smartnode.GetMinipoolPerformancePath(index, true)
-	err = os.WriteFile(minipoolPerformancePath, minipoolPerformanceBytes, 0644)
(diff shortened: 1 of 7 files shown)


rocketpool/watchtower/generate-rewards-tree.go
jshufro, line 1: This file runs in watchtower on non-odao nodes who wish to generate the tree instead of downloading it.

As such it simply saves the files and doesn't calculate cids

rocketpool/watchtower/submit-rewards-tree-rolling.go
```diff
@@ -391,39 +389,35 @@ func (t *submitRewardsTree_Rolling) getTrueRewardsIntervalSubmissionSlot(targetS
 }
 
 // Checks to see if an existing rewards file is still valid and whether or not it should be regenerated or just resubmitted
-func (t *submitRewardsTree_Rolling) isExistingFileValid(rewardsTreePath string, intervalsPassed uint64, nodeAddress common.Address, isInOdao bool) (rprewards.IRewardsFile, []byte, bool, bool) {
+func (t *submitRewardsTree_Rolling) isExistingFileValid(rewardIndex uint64, intervalsPassed uint64, nodeAddress common.Address, isInOdao bool) (*rprewards.LocalRewardsFile, bool, bool) {
```
jshufro, line 392: The new signature simply returns a LocalFile[IRewardsFIle] instead of the byteslice and IRewardsFile, since the new type encapsulates both.

I also replaced rewardsTreePath with rewardIndex to make it a bit more generic.

rocketpool/watchtower/submit-rewards-tree-rolling.go
```diff
@@ -391,39 +389,35 @@ func (t *submitRewardsTree_Rolling) getTrueRewardsIntervalSubmissionSlot(targetS
...
-	if err != nil {
-		t.log.Printlnf("%s WARNING: failed to deserialize %s: %s; regenerating file...\n", t.logPrefix, rewardsTreePath, err.Error())
-		return nil, nil, false, true
-	}
+	proofWrapper := localRewardsFile.Repr()
 	header := proofWrapper.GetHeader()
 
 	if isInOdao {
 		// Get the CID for it
-		cid, err := rprewards.GetCidForRewardsFile(proofWrapper, filename)
```
jshufro, line 423: this line is the buggy one.

`proofWrapper` is decompressed, `GetCidForRewardsFile` ultimately calls `SingleFileDirIPFSCid` which does compress it eventually, but doesn't add .zst to filename before calculating the cid:
https://github.com/rocket-pool/smartnode/pull/432/files#diff-51c4c70c13d682999d12284148b72676b8fe2672757f756401d049f950c18793L83

because `filename` here is the uncompressed file name, the cid would be incorrect.

PR #433, opened 2024-01-13
Don't omit totalNodeWeight if not set. Add a test to cement CID calculation.

jshufro:


shared/services/rewards/types.go
```diff
@@ -73,7 +73,7 @@ type TotalRewards struct {
 	TotalSmoothingPoolEth        *QuotedBigInt `json:"totalSmoothingPoolEth"`
 	PoolStakerSmoothingPoolEth   *QuotedBigInt `json:"poolStakerSmoothingPoolEth"`
 	NodeOperatorSmoothingPoolEth *QuotedBigInt `json:"nodeOperatorSmoothingPoolEth"`
-	TotalNodeWeight              *QuotedBigInt `json:"totalNodeWeighti,omitempty"`
```
jshufro: lol i blame vim

nice catch, that would have been bad

PR #434, opened 2024-01-18
Use the Go version of the Nethermind prune starter, replacing the dll

0xfornax:
To avoid .NET version conflicts as seen with version 1.11.5 we're replacing the dll with a Golang application that handles the calls to start a NM prune.

We're also making it unnecessary to restart Nethermind to prune.


rocketpool-cli/service/service.go
```diff
@@ -1055,42 +1055,43 @@ func pruneExecutionClient(c *cli.Context) error {
...
-	if err != nil {
-		return fmt.Errorf("Error getting execution client volume name: %w", err)
-	}
+	if selectedEc == cfgtypes.ExecutionClient_Nethermind {
+		// Restarting NM is not needed anymore
+		err = rp.RunNethermindPruneStarter()
+		if err != nil {
+			return fmt.Errorf("Error starting Nethermind prune starter: %w", err)
+		}
+	} else {
```
jshufro, line 1064:
```suggestion
		}
		return nil
	}
```

rocketpool-cli/service/service.go
```diff
@@ -1055,42 +1055,43 @@ func pruneExecutionClient(c *cli.Context) error {
...
+		fmt.Printf("Restarting %s...\n", executionContainerName)
+		result, err = rp.StartContainer(executionContainerName)
 		if err != nil {
-			return fmt.Errorf("Error starting Nethermind prune starter: %w", err)
+			return fmt.Errorf("Error starting main execution client: %w", err)
+		}
+		if result != executionContainerName {
+			return fmt.Errorf("Unexpected output while starting main execution client: %s", result)
 		}
 	}
```
jshufro, line 1096:
```suggestion
	}
```

rocketpool/nm-prune-starter/nm-prune-starter-linux-amd64
jshufro, line 1: are we not publishing these to docker-hub? i feel like the binaries should be .gitignore and not revisioned. native mode users don't need prune provisioner (they manage the nm install), and docker users can get it from dockerhub like all the other containers

shared/services/rocketpool/client.go
```diff
@@ -784,9 +783,10 @@ func (c *Client) RunPruneProvisioner(container string, volume string, image stri
 
 }
 
-// Runs the prune provisioner
-func (c *Client) RunNethermindPruneStarter(container string) error {
-	cmd := fmt.Sprintf("docker exec %s %s %s", container, nethermindPruneStarterCommand, nethermindAdminUrl)
+// Executes a Go program that triggers NM pruning
+func (c *Client) RunNethermindPruneStarter() error {
+	cmd := fmt.Sprintf(`docker run --rm --network container:rocketpool_eth1 rocketpool/nm-prune-starter %s`, nethermindAdminUrl)
```
jshufro, line 788: first, time to make this into a function: https://github.com/rocket-pool/smartnode/blob/d30dd1ff2097c5430d82492ab2199d4f41c7fa7e/rocketpool-cli/service/service.go#L1028

second,

```suggestion
	cmd := fmt.Sprintf(`docker run --rm --network container:%s rocketpool/nm-prune-starter %s`, newFunction(), nethermindAdminUrl)
```

jshufro (changes requested):

rocketpool/nm-prune-starter/nm-prune-starter-linux-amd64
jshufro: sounds good. as long as the dockerfile is in situ, i don't care about monorepo vs manyrepo

jshufro: i assume you're also removing the binaries, and they were only versioned here for the dockerfile to work.

when you move the dockerfile you should do a multipart build instead of revisioning compiled code, which reduces automation

shared/services/rocketpool/client.go
jshufro: one more thing: consider adding a `--name` flag to the container so it is easy to identify if it gets stuck or hangs indefinitely or whatever.

something like `--name %s` which is populated respecting the project name prefix and some constant container name, like `nm-prune-starter`

jshufro:
Bleh.

```golang
// Gets the prefix specified for Rocket Pool's Docker containers
func getContainerPrefix(rp *rocketpool.Client) (string, error)
```
should be 
```golang
// Gets the prefix specified for Rocket Pool's Docker containers
func (rp *rocketpoolClient) getContainerPrefix() (string, error)
```

and make
```golang
func getExecutionContainerName(prefix string) string
```
into
```golang
func (rp *rocketpoolClient) GetExecutionContainerName() (string, error)
```

passing `prefix` as a string is too ambiguous- you can't immediately tell whether it contains the separator `_`.

Consider doing all this as a separate branch, though, because it's going to require changes elsewhere

0xfornax:
Well, prefix + ContainerSuffix is everywhere so I'm leaving it for now. But you got the

> ```go
> (rp *rocketpoolClient) getContainerPrefix()
> ```

jshufro:
> Well, prefix + ContainerSuffix is everywhere

yeah, that's the problem 😆

rocketpool-cli/service/service.go
```diff
@@ -1055,6 +1043,16 @@ func pruneExecutionClient(c *cli.Context) error {
...
 	fmt.Printf("Your disk has %s free, which is enough to prune.\n", freeSpaceHuman)
 
+	if selectedEc == cfgtypes.ExecutionClient_Nethermind {
+		// Restarting NM is not needed anymore
+		err = rp.RunNethermindPruneStarter(executionContainerName)
+		if err != nil {
+			return fmt.Errorf("Error starting Nethermind prune starter: %w", err)
+		}
+		fmt.Printf("\nDone! Your main execution client is now pruning. You can follow its progress with `rocketpool service logs eth1`.\n")
+		fmt.Printf("%sNOTE: While pruning, you **cannot** interrupt the client (e.g. by restarting) or you risk corrupting the database!\nYou must let it run to completion!%s\n", colorYellow, colorReset)
```
jshufro, line 1053: print these in `RunNethermindPruneStarter` on success.

The messaging here is slightly different than for geth, so it makes sense to copy it into the nethermind-specific function

rocketpool-cli/service/service.go
```diff
@@ -1055,6 +1043,16 @@ func pruneExecutionClient(c *cli.Context) error {
 
 	fmt.Printf("Your disk has %s free, which is enough to prune.\n", freeSpaceHuman)
 
+	if selectedEc == cfgtypes.ExecutionClient_Nethermind {
+		// Restarting NM is not needed anymore
+		err = rp.RunNethermindPruneStarter(executionContainerName)
+		if err != nil {
+			return fmt.Errorf("Error starting Nethermind prune starter: %w", err)
+		}
+		fmt.Printf("\nDone! Your main execution client is now pruning. You can follow its progress with `rocketpool service logs eth1`.\n")
```
jshufro, line 1052: split this into:
```golang
fmt.Println()
fmt.Println("Your main execution client is now pruning. You can follow its progress with `rocketpool service logs eth1`.")
```

there's no need for printf with a single string, and separate printlns makes the code read closer to the output, which improves readability

rocketpool-cli/service/service.go
jshufro, line 1053: recommend changing into:
```golang
fmt.Println(colorYellow + "NOTE: While pruning, you **cannot** interrupt the client (e.g. by restarting) or you risk corrupting the database!\nYou must let it run to completion!" + colorReset)
```

the compiler should be smart enough to see that colorYellow, colorReset are `const` and do this automatically, but if it isn't, this is technically much quicker, since the string concatenation will happen at comptime

jshufro (changes requested):

jshufro (approved):

PR #447, opened 2024-02-15
Fix voluntary exit signatures

0xfornax:
Fix the exit signatures as they were failing after the Dencun upgrade on Holesky.
Always using the CAPELLA_FORK_VERSION when computing the domain for the voluntary exit signature.
Domain() function was deprecated so it was replaced by ComputeDomain()

### DIFF
--- a/rocketpool/api/minipool/change-withdrawal-creds.go
+++ b/rocketpool/api/minipool/change-withdrawal-creds.go
@@ -198,7 +198,7 @@ func changeWithdrawalCreds(c *cli.Context, minipoolAddress common.Address, mnemo
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


jshufro:
will this work on mainnet? what happens when the CAPELLA_FORK_VERSION field is null, eg on a pre-capella testnet?

0xfornax:
> will this work on mainnet? what happens when the CAPELLA_FORK_VERSION field is null, eg on a pre-capella testnet?

Yes, on mainnet the signature uses the current fork version (which is also Capella).
I don't think it's worth trying to support past forks and also don't see the benefits of running such forks on testnets as they don't allow us to replicate the production environment.

Issue #186, opened 2022-01-20
node-collector.go's `Collect` function all-or-nothing


jshufro:
`Collect` spins up a bunch of goroutines in a waitgroup to gather stats, then waits for them, exiting early if any one fails.

Subsequently it publishes them all synchronously. This means if any one metric can't be computed due to some transient error, none are.

This decision seems to be a result of the fact some metrics are derived from other stats, and those stats need to be observed before the derived metrics can be published.

One option would be to compute all the derived metrics in grafana, and only publish raw metrics to prometheus. Alternatively, the raw metrics could asynchronously produce their results on the output channel (which is thread-safe), and the derived metrics could synchronously produce their results on the channel iff all the raw metrics succeed.

PR #449, opened 2024-02-22
feat: smartnode notification functionality for bounty BA022308 (PR 1 of 3)

activescott:
This is an implementation of [Smartnode Notification Functionality bounty BA022308](https://longforwisdom.github.io/rp-bounty/BA022308).
The approach is based on an ongoing [discord discussion thread](https://discord.com/channels/1109303903767507016/1204627769724637275) and @jshufro was kind enough to provide [some implementation guidance](https://gist.github.com/jshufro/e86b61fcd131d4c99938680934d854eb).

This PR is related to:
- Templates and updates in  rocket-pool/smartnode-install#119 
- docs are in rocket-pool/docs.rocketpool.net#73

Below is a screenshot of the TUI: 
Screenshot 2024-03-04 at 16 34 13

Below is a screenshot of what some alerts look like with a configured Discord webhook URL:
Screenshot 2024-02-21 at 15 35 46

Below is a screenshot of `rocketpool node status` output:
Screenshot 2024-02-23 at 12 47 31

## Todo
This is a draft to get some initial feedback. What's included and remaining todos are (across both of these PRs):

- [x] Alertmanager container managed by rocketpool-cli
- [x] Support for Discord Notifications (Configuration and docs in TUI)
- [x] Support for notifications to all Alertmanager receivers (email, msteams, opsgenie, pagerduty, pushover, slack, sns, telegram, victorops, webhook, wechat, webex) using yml configuration.
- [x] [Alertmanager client](https://prometheus.io/docs/alerting/latest/clients/) implementation to implement remaining ephemeral alerts. NOTE: Alertmanager client will handle things such as notification rate limiting (if an alert is sent by the node, resolves, and resent rapidly) and auto-resolving (e.g. once the node stop sending an alert to alertmanager, it will be automatically resolved based on a configurable time-based threshold which can be set per alert and globally).
- [x] Configured alerts from bounty:
  - [x] 1. Client sync complete
  - [x] 2. Client(s) lost sync (`ClientSyncStatusBeacon`, `ClientSyncStatusExecution`, implemented via new prometheus metrics named `rocketpool_node_sync_progress` using the same sync status calculation that `rocketpool node sync` uses).
  - [x] 3. ** Disk free space is running low (say, 15% free remaining) and the user should consider pruning if using Geth or Nethermind _(`LowDiskSpaceWarning`, `LowDiskSpaceCritical` provide a warning level and a critical level both with thresholds)_
  - [x] 4. node automatically staked a minipool, or attempted to (success or failure)
  - [x] 5. node automatically promoted a vacant minipool, or attempted to (success or failure)
  - [x] 6. node automatically reduced a minipool’s bond, or attempted to (success or failure)
  - [x] 7. node detected a fee recipient change
  - [x] 8. node automatically distributed a minipool’s balance
  - [x] 9.  You have a block proposal this epoch (`UpcomingProposal` using the same metric shown on the grafana dashboard)
  - [x] 10. You submitted a block proposal
  - [x] 11. You are scheduled for the next sync committee (`UpcomingSyncCommittee` using the same metric shown on the grafana dashboard)
  - [x] 12. You have entered a sync committee (`ActiveSyncCommittee` using the same metric shown on the grafana dashboard)
  - [x] 13. There is a new Smartnode update available (`RPUpdatesAvailable`)

- Additional features beyond what is mentioned in bounty:
  - [x] support for user-defined rules in `/alerting/rules/*.yml` using [Prometheus rule configuration](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/).
  - [x] Alert: Rocket Pool OS Updates Available (`OSUpdatesAvailable`)
  - [x] `rocketpool node status` shows current actively firing alerts
  - [ ] (optional) Alert: Staked RPL collateral falls below minimum level to claim RPL rewards (10% of borrowed ETH)

- [x] Basic TUI: Ability to configure alertmanager port & a discord webhook receiver/notification
- [x] Advanced TUI: Ability to enable/disable specific receivers and groups of alerts for specific receivers + new TUI page
  - [x] Add the 7 custom alerts that are fired from shared/services/alerting/alerting.go to the TUI configuration for enablement.

- Native Mode users support:
  - [-] ~~instructions on creating and managing a systemd service for Native Mode users~~ _I'd like to recommend dropping this. The implementation is heavily related to setting up the Prometheus/Grafana stack and the current native mod docs direct the user to the docker section to set those up. Largely this is going to be installing Prometheus and Alertmanager and configuring them. If someone is advanced enough to do that I doubt they'll bother with our docs._
  - Anything else for native mode users? Does something need disabled in docker config or something??
- [x] a pull request to our documentation guides repository (GitHub - rocket-pool/docs.rocketpool.net: Rocket Pool Documentation & Guide Hub) with complete and thorough documentation describing its configuration and usage for Docker, Hybrid, and Native Mode users alike. – _see rocket-pool/docs.rocketpool.net#73_

## Open Questions

- [x] There is "first-class UI" configuration in TUI for the basics of alert manager including a Discord Webhook URL for notifications and ports similar to other containers. Currently each alert is configured in an Alertmanager rules config file with configurable thresholds at  [`~/.rocketpool/alerting/rules/default.yml`](https://github.com/rocket-pool/smartnode-install/pull/119/files#diff-9a6b401952feb43bb3597eb0d95eea87355eeb67dab52185e5fde35132fc19f5). I can add each alert to the TUI, but this is of questionable value in IMHO considering the documentation and flexibility of the alertmanager rules file. Does the GMC want each notification in the TUI or is the rules yml acceptable? 
  - NOTE: The bounty mentions _Notifications must be configurable, meaning they need to have first-class support as a dedicated page in the service config TUI with parameters to enable and disable each notification, and for adjustable thresholds where appropriate._
  - As discussed [below](https://github.com/rocket-pool/smartnode/pull/449#issuecomment-1960559638) we will implement TUI for this. Task exists above.


activescott:
@jshufro Will appreciate your review of this before I move onto the additional "harder" alerts. Specifically looking for any feedback on technical direction/factoring and the remaining Todos above which I believe, once completed, will complete the bounty. 

Feel free to tag others from GMC or dev team that might be appropriate to weigh in here.

rocketpool/node/collectors/node-collector.go
```diff
@@ -339,6 +352,33 @@ func (collector *NodeCollector) Collect(channel chan<- prometheus.Metric) {
...
+		syncStatus, err := collector.bc.GetSyncStatus()
+
+		if err != nil {
+			// don't return an error because then the metric isn't emitted at all, just let the Progress be 0
+			fmt.Printf("error getting beacon chain sync status: %w", err)
+		} else {
+			progress = syncStatus.Progress
+		}
+
+		channel <- prometheus.MustNewConstMetric(
```
activescott, line 368: FYI: Note that the channel is used here asynchronously rather than at the bottom of the outer function because this function aborts when the client's aren't synced (since most of the other calculations here require a synced client). I did some testing and this allows these metrics to be emitted even when the client isn't synced while the other ones are not emitted at all on the scrape endpoint until the clients are synced.

shared/services/config/alertmanager-config.go
```diff
@@ -0,0 +1,132 @@
...
+		&cfg.ContainerTag,
+		&cfg.DiscordWebhookURL,
+	}
+}
+
+func (cfg *AlertmanagerConfig) GetConfigTitle() string {
+	return cfg.Title
+}
+
+// Used by text/template to format alertmanager.yml
```
activescott, line 104: `GetOpenPorts` and `UpdateConfigurationFile` have a very similar implementation to equivelents elsewhere but I felt the code was cleaner and better factored by putting them here instead of sprinkling the "alerting concern" throughout the codebase so much.

jshufro:
> @jshufro Will appreciate your review of this before I move onto the additional "harder" alerts. Specifically looking for any feedback on technical direction/factoring and the remaining Todos above which I believe, once completed, will complete the bounty.
> 
> Feel free to tag others from GMC or dev team that might be appropriate to weigh in here.

Will give this a read over tomorrow!

rocketpool-cli/service/config/settings-metrics.go
```diff
@@ -23,6 +23,7 @@ type MetricsConfigPage struct {
 	watchtowerMetricsPortBox   *parameterizedFormItem
 	grafanaItems               []*parameterizedFormItem
 	prometheusItems            []*parameterizedFormItem
+	alertmanagerItems					 []*parameterizedFormItem
```
jshufro, line 26: we use goimports for formatting: https://pkg.go.dev/golang.org/x/tools/cmd/goimports

rocketpool/node/collectors/node-collector.go
jshufro: Yep, noted in: https://github.com/rocket-pool/smartnode/issues/186

Just hasn't been a priority to make a better dependency graph here.

Add comments to the code?

rocketpool/node/collectors/node-collector.go
```diff
@@ -339,6 +352,33 @@ func (collector *NodeCollector) Collect(channel chan<- prometheus.Metric) {
...
 	})
 
+	// get the beacon client sync status:
+	wg.Go(func() error {
+		progress := float64(0)
+		
+		syncStatus, err := collector.bc.GetSyncStatus()
+
+		if err != nil {
+			// don't return an error because then the metric isn't emitted at all, just let the Progress be 0
```
jshufro, line 362: I worry that this will be very noisy on startup.

If rocketpool_node and the beacon node are restarted, then prometheus restarts milliseconds later, it will scrape rocketpool_node potentially before the beacon node is accepting connections, resulting in a timeout or tcp or http error here.

I don't think we care to alert on sync status when there is an error, and omitting the metric is probably best. Alerting when the beacon node is up, replying, and sync is <1 is much more _specific_, while doing it the way you've handled it here is much more _sensitive_.

I think specific wins out in this case because i anticipate other alerts to fire if the bn is offline

shared/services/config/alertmanager-config.go
jshufro: it's on my mental todo list to move the text/template specific functions closer to the values they scan.

jshufro (commented):
very strong start.

I _do_ think we want more flexibility in the TUI, at a minimum the ability to enable/disable specific receivers and groups of alerts for specific receivers.

You don't want node operators editing yaml themselves... it's not pretty.

rocketpool/node/collectors/node-collector.go
activescott: Ha. I didn't see that issue. FYI Async publishing seemed to work just fine here. I tested by doing things like shutting down the BC and scrape /metrics the endpoint with wget rapidly and all seemed to behave fine.

rocketpool-cli/service/config/settings-metrics.go
activescott: ack. I think I fixed this in vscode now.

rocketpool/node/collectors/node-collector.go
activescott: I'm open here. Just for some context:

I've tested this (and just did again) by shutting down the eth2 container entirely and scraping the metrics endpoint rapidly and all works fine. There is now **three** log messages instead of two on _each scrape_ but that's all (the others are `[Node Collector] Error getting beacon chain head: no Beacon clients were ready` and `[Beacon Collector]`). I could remove this log message?

If you're worried about noisy alerts we can control that in the prometheus rule configuration for the alert so that it won't fire unless it's offline for a specified period. 

So with that in mind, what do you prefer to be done here specifically?

activescott:
> I _do_ think we want more flexibility in the TUI, at a minimum the ability to enable/disable specific receivers and groups of alerts for specific receivers.
> 
> You don't want node operators editing yaml themselves... it's not pretty.

Okay how about I add a TUI entry for each "built in" alert with the choices "Do not alert" and "Discord" for now. I can use that to template the default rules file. Then other receivers can be added to the choices later (as well as an "All" I think?). Ok?

rocketpool/node/collectors/node-collector.go
jshufro: log messages are not a concern (unless we're talking a dozen). If the alertmanager `for` configs make restarts sufficiently quiet, I don't mind it as-is.

jshufro:
> > I _do_ think we want more flexibility in the TUI, at a minimum the ability to enable/disable specific receivers and groups of alerts for specific receivers.
> > You don't want node operators editing yaml themselves... it's not pretty.
> 
> Okay how about I add a TUI entry for each "built in" alert with the choices "Do not alert" and "Discord" for now. I can use that to template the default rules file. Then other receivers can be added to the choices later (as well as an "All" I think?). Ok?

Sounds like a good approach. You might want to make a new TUI page sooner rather than later. It sounds like it's going to get 'busy' and having an alerts/notifications dedicated page will buy you more real-estate.

I'm kind of envisioning a checkbox style grid:

```
+-----------------+----------+----------+-----------+
|                 | discord  | email    | pagerduty |
| proposal alerts |    x     |     x    |           |
| hardware alerts |          |    x     |    x      |
| slashing alerts |          |          |      x    |
| ...             |          |          |           |
+-----------------+----------+----------+-----------+
```

something like this. Ideally hiding receivers that aren't configured. It's a big ask, but anything you can do directionally towards this would be helpful.

rocketpool/node/collectors/node-collector.go
```diff
@@ -339,6 +352,33 @@ func (collector *NodeCollector) Collect(channel chan<- prometheus.Metric) {
...
 
+	// get the beacon client sync status:
+	wg.Go(func() error {
+		progress := float64(0)
+
+		syncStatus, err := collector.bc.GetSyncStatus()
+
+		if err != nil {
+			// don't return an error because then the metric isn't emitted at all, just let the Progress be 0
+			fmt.Printf("error getting beacon chain sync status: %w", err)
```
jshufro, line 363: revisiting this i think `return nil` here (after printing the error) would be good:

On startup, we don't report sync progress _until_ the beacon client is up.
After we get sync progress at least once, the last know sync progress will continue to be reported.

This way, the endpoint stays response, but we also use slightly more accurate signals

rocketpool/node/distribute-minipools.go
```diff
@@ -163,8 +164,11 @@ func (t *distributeMinipools) run(state *state.NetworkState) error {
 	for _, mpd := range minipools {
 		success, err := t.distributeMinipool(mpd, opts)
 		if err != nil {
+			alerting.AlertMinipoolBalanceDistributed(t.cfg, mpd.MinipoolAddress, false)
 			t.log.Println(fmt.Errorf("Could not distribute balance of minipool %s: %w", mpd.MinipoolAddress.Hex(), err))
 			return err
+		} else {
+			alerting.AlertMinipoolBalanceDistributed(t.cfg, mpd.MinipoolAddress, true)
```
jshufro, line 171:
```suggestion
		success, err := t.distributeMinipool(mpd, opts)
		alerting.AlertMinipoolBalanceDistributed(t.cfg, mpd.MinipoolAddress, err == nil)
		if err != nil {
			t.log.Println(fmt.Errorf("Could not distribute balance of minipool %s: %w", mpd.MinipoolAddress.Hex(), err))
			return err
		} 
```

incredibly minor semantics improvement - easier to see alerting is unconditional at a glance.

rocketpool/node/manage-fee-recipient.go
```diff
@@ -120,12 +121,15 @@ func (m *manageFeeRecipient) run(state *state.NetworkState) error {
...
 		m.log.Println("Shutting down the validator client for safety to prevent you from being penalized...")
+		alerting.AlertFeeRecipientChanged(m.cfg, correctFeeRecipient, false)
 
 		err = validator.StopValidator(m.cfg, m.bc, &m.log, m.d)
 		if err != nil {
 			return fmt.Errorf("error stopping validator client: %w", err)
 		}
 		return nil
+	} else {
+		alerting.AlertFeeRecipientChanged(m.cfg, correctFeeRecipient, true)
```
jshufro, line 132: Same deal here- do the alerting between UpdateFeeRecipientFile and the conditional

rocketpool/node/node.go
```diff
@@ -154,23 +155,41 @@ func run(c *cli.Context) error {
 
 	// Run task loop
 	go func() {
+		// we assume clients are synced on startup so that we don't send unnecessary alerts
+		wasExecutionClientSynced := true
+		wasBeaconClientSynced := true
 		for {
```
jshufro, line 161: you found my favorite tech debt!

rocketpool/node/promote-minipools.go
```diff
@@ -131,7 +132,10 @@ func (t *promoteMinipools) run(state *state.NetworkState) error {
 		_, err := t.promoteMinipool(mpd, opts)
 		if err != nil {
 			t.log.Println(fmt.Errorf("Could not promote minipool %s: %w", mpd.MinipoolAddress.Hex(), err))
+			alerting.AlertMinipoolPromoted(t.cfg, mpd.MinipoolAddress, false)
 			return err
+		} else {
+			alerting.AlertMinipoolPromoted(t.cfg, mpd.MinipoolAddress, false)
```
jshufro, line 138: (here too)

rocketpool/node/reduce-bonds.go
```diff
@@ -176,7 +177,10 @@ func (t *reduceBonds) run(state *state.NetworkState) error {
 		success, err := t.reduceBond(mp, windowStart, windowLength, latestBlockTime, opts)
 		if err != nil {
 			t.log.Println(fmt.Errorf("could not reduce bond for minipool %s: %w", mp.MinipoolAddress.Hex(), err))
+			alerting.AlertMinipoolBondReduced(t.cfg, mp.MinipoolAddress, false)
 			return err
+		} else {
+			alerting.AlertMinipoolBondReduced(t.cfg, mp.MinipoolAddress, true)
```
jshufro, line 183: (und here)

rocketpool/node/stake-prelaunch-minipools.go
```diff
@@ -149,6 +150,9 @@ func (t *stakePrelaunchMinipools) run(state *state.NetworkState) error {
 		}
 		if success {
 			successCount++
+			alerting.AlertMinipoolStaked(t.cfg, mpd.MinipoolAddress, true)
+		} else {
+			alerting.AlertMinipoolStaked(t.cfg, mpd.MinipoolAddress, false)
```
jshufro, line 155: (und here)

shared/services/alerting/alerting.go
```diff
@@ -0,0 +1,259 @@
...
+	DefaultEndsAtDurationForSeverityCritical = time.Minute * 60
+)
+
+// fetches the current alerts directly the alertmanager container/application's API.
+// If alerting/metrics are disabled, this function returns an empty array.
+func FetchAlerts(cfg *config.RocketPoolConfig) ([]*models.GettableAlert, error) {
+	// NOTE: don't log to stdout here since this method is on the "api" path and all stdout is parsed as a json "api" response.
+	if !isAlertingEnabled(cfg) {
+		// metrics are disabled, so no alerts will be fetched.
+		return make([]*models.GettableAlert, 0), nil
```
jshufro, line 27: ideally we should be able to return `nil, nil` to indicate success, no alerts

i'm guessing you segfaulted somewhere? shouldn't happen if you `len` check first, or only use `range` which is nil-safe

shared/services/requirements.go
```diff
@@ -523,6 +526,11 @@ func waitBeaconClientSynced(c *cli.Context, verbose bool, timeout int64) (bool,
 	// Get BC status refresh time
 	bcRefreshTime := startTime
 
+	cfg, err := getConfig(c)
+	if err != nil {
+		return false, err
```
jshufro, line 531: ideally fmt.Errorf() wrap this error with %w and a little message about error getting config while checking bc

rocketpool/node/node.go
activescott: ha. I found it odd that it waits since the system has to be able to detail with out-of-sync clients at all times anyway. If you want I can clean this up but I'd prefer to do it in a separate PR after this one just to keep them independent and simple.

shared/services/alerting/alerting.go
activescott: Purely my ignorance of go. I actually did learn about this nil-safe iteration at some point but it was a bit after I wrote this. In other languages I always prefer empty collections instead of null/nil but I guess go has taken that classic bug into account.
Fixed.

rocketpool/node/collectors/beacon-collector.go
```diff
@@ -178,7 +199,35 @@ func (collector *BeaconCollector) Collect(channel chan<- prometheus.Metric) {
...
+		collector.recentProposals, prometheus.GaugeValue, recentProposalCount)
+}
+
+func (collector *BeaconCollector) getProposedBlockCount(validatorIndices []string, head beacon.BeaconHead, slotsPerEpoch uint64) (float64, error) {
+	// prepare for quick lookups in event of many validators:
+	indexLookup := make(map[string]string, len(validatorIndices))
+	for _, index := range validatorIndices {
+		indexLookup[index] = index
+	}
+	latestSlot := head.FinalizedEpoch*slotsPerEpoch + (slotsPerEpoch - 1)
```
activescott, line 212: I ended up using finalized here just seemed the most straightforward and easiest to explain. Who knows baseball but us yanks anyway? :) If different values for choosing the "latest" slot are preferred here, just suggest them or commit them :)

rocketpool/node/node.go
jshufro: nah don't sweat it. i think joe might have fixed it on a branch.

it's on my mental backlog to replace this for loop with a time.Ticker-based abstraction, if he didn't, though

shared/services/bc-manager.go
```diff
@@ -161,6 +161,16 @@ func (m *BeaconClientManager) GetBeaconBlock(blockId string) (beacon.BeaconBlock
 	return result1.(beacon.BeaconBlock), result2.(bool), nil
 }
 
+func (m *BeaconClientManager) GetBeaconBlockHeader(blockId string) (beacon.BeaconBlockHeader, bool, error) {
```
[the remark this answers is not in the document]
jshufro: agreed

rocketpool/node/collectors/beacon-collector.go
```diff
@@ -178,7 +199,35 @@ func (collector *BeaconCollector) Collect(channel chan<- prometheus.Metric) {
...
+		if err != nil {
+			collector.logError(fmt.Errorf("error getting beacon block: %w", err))
+			continue
+		}
+		if !hasBlock {
+			continue
+		}
+		if _, ok := indexLookup[block.ProposerIndex]; ok {
+			proposedBlockCount++
+		}
```
jshufro, line 228:
```suggestion
		if _, ok := indexLookup[block.ProposerIndex]; !ok {
			continue
		}
		proposedBlockCount++
```

incredibly nitpicky, but i think:
```
for {
  if thing1 {
    continue
  }
  if thing2 {
    continue
  }
  if thing3 {
    continue
  }
  success()
}
```

is a more semantic expression here

rocketpool-cli/service/config/standard-layout.go
```diff
@@ -205,3 +205,23 @@ func (layout *standardLayout) mapParameterizedFormItems(params ...*parameterized
 		layout.parameters[param.item] = param
 	}
 }
+
+// Sets up a handler to return to the specified homePage when the user presses escape on the layout.
+func (layout *standardLayout) setupEscapeReturnHomeHandler(md *mainDisplay, homePage *page) {
```
activescott, line 210: FYI: I found 10 copies of this code throughout the codebase and replaced them all with this function.

activescott:
@jshufro When ya'll are back this PR and the corresponding PR in rocket-pool/smartnode-install#119 are ready to have a "final" review and some testing. I believe this is now feature complete, I just need to get a PR together for docs.

activescott:
I believe that the combination of the 3 PRs mentioned above now meets the requirements of the bounty. While I think there are things that could be added (e.g. more delivery channels, more alerts, more TUI configuration options) this has become a significant body of work and a meaningful improvement for node operators so I'd like to follow up additional functionality in subsequent milestones.

I do have some notes above pointing out some things that went beyond the requirements in the bounty and any somewhat "grey areas" (e.g. native mode) in the interest of transparency. Happy to discuss anything at all.

What else can I do to move this forward?

go.mod
```diff
@@ -14,8 +14,13 @@ require (
 	github.com/ferranbt/fastssz v0.1.3
 	github.com/gdamore/tcell/v2 v2.6.0
 	github.com/glendc/go-external-ip v0.1.0
+	github.com/go-openapi/errors v0.21.0
```
activescott, line 17: FYI: I needed the openapi dependencies for the client to call alertmanager. The other version bumps were done by go. I didn't notice any problems with those bumps but they were really just done by the `go mod...` command (get or tidy, I forget which).

jshufro:
> What else can I do to move this forward?

I hope/think it's only stalled because of conference travel. @0xfornax should be able to approve and merge. @jclapis is the final verdict on whether the bounty is fulfilled but he's on paternity leave so we may have to ask the GMC if they're willing to take fornax/my word instead.

I just got back from Denver so I'll test a bit and do another review pass. @shfryn fyi

0xfornax:
Greetings @activescott and @jshufro! I've started testing/reviewing the 3 PRs.

rocketpool-cli/node/status.go
```diff
@@ -312,6 +312,19 @@ func getStatus(c *cli.Context) error {
 		fmt.Println("The node is not registered with Rocket Pool.")
 	}
 
+	// Alerts
+	if cfg.EnableMetrics.Value == true {
+		// only print alerts if enabled; to avoid misleading the user to thinking everything is fine (since we really don't know).
+		fmt.Printf("\n%s=== Alerts ===%s\n", colorGreen, colorReset)
+		for _, alert := range status.Alerts {
+			fmt.Println(alert.ColorString())
```
0xfornax, line 320: I'm worried this pollutes the status page. Would rather see somethings like "There are X alerts enabled", and then a specific command to list them.

jshufro: pretty sure it only prints alerts that are actively firing.

we should print the `=== Alerts ===` conditionally on there being any firing alerts though

activescott: I removed alerts section when no alerts are firing and limited it to 3 alerts in ecb830c5cd66e78fa1daabc10f67e06763882259

rocketpool-cli/node/status.go
```diff
@@ -312,6 +312,22 @@ func getStatus(c *cli.Context) error {
...
+	if cfg.EnableMetrics.Value == true && len(status.Alerts) > 0 {
+		// only print alerts if enabled; to avoid misleading the user to thinking everything is fine (since we really don't know).
+		fmt.Printf("\n%s=== Alerts ===%s\n", colorGreen, colorReset)
+		const maxItems = 3
+		for i, alert := range status.Alerts {
+			fmt.Println(alert.ColorString())
+			if i == maxItems-1 {
+				break
+			}
+		}
```
jshufro, line 325:
```suggestion
		const maxItems = 3
		for i, alert := range status.Alerts {
			if i == maxItems {
				break
			}
			fmt.Println(alert.ColorString())
		}
```

upside- no need to reason about off-by-one errors
downside- one extra iteration if the compiler doesn't catch and optimize it.

this is the softest suggestion i think i've ever given though, so feel free to ignore

Issue #450, opened 2024-03-02
`rocketpool node stake-rpl` reads incorrect selection


jduffey:
Reproduction steps:
- Run `rocketpool node stake-rpl` in order to select how much RPL to stake.
- Example output:
```
Please choose an amount of RPL to stake:
1: The minimum minipool stake amount for an 8-ETH minipool (244.921431 RPL)?
2: The minimum minipool stake amount for a 16-ETH minipool (163.280954 RPL)?
3: Your entire RPL balance (265.331549 RPL)?
4: A custom amount
```
- Select option 3 to stake the entire RPL balance
- Observe that you're presented with the prompt, "Please enter an amount of RPL to stake:"
- Cancel the interaction (e.g. Ctrl + C) and run the command again
- Select option 4 to stake a custom amount
- Observe that you're not presented with the option to select the amount of RPL and instead are prompted to select your gas price, as though the RPL amount has already been decided

I think the issue lies in `rocketpool-cli/node/stake-rpl.go`, line 210, where `case 3` should be `case 2`; context:
```
// Prompt for amount option
amountOptions := []string{
	fmt.Sprintf("The minimum minipool stake amount for an 8-ETH minipool (%.6f RPL)?", math.RoundUp(eth.WeiToEth(minAmount8), 6)),
	fmt.Sprintf("The minimum minipool stake amount for a 16-ETH minipool (%.6f RPL)?", math.RoundUp(eth.WeiToEth(minAmount16), 6)),
	fmt.Sprintf("Your entire RPL balance (%.6f RPL)?", math.RoundDown(eth.WeiToEth(&rplBalance), 6)),
	"A custom amount",
}
selected, _ := cliutils.Select("Please choose an amount of RPL to stake:", amountOptions)
switch selected {
case 0:
	amountWei = minAmount8
case 1:
	amountWei = minAmount16
case 3:
	amountWei = &rplBalance
}
```

PR #451, opened 2024-03-02
Issue #450 - Bugfix when smartnode is reading user sele…

jduffey:
Bugfix when smartnode is reading user selection from `rocketpool node stake-rpl`

Reference: https://github.com/rocket-pool/smartnode/issues/450


0xfornax (approved):
Thanks for the contribution! lgtm

PR #459, opened 2024-03-11
feat: Migrate networks to yml file (1 of 2)

### LINKED ISSUE #438: Networks should not be hard-coded
Smartnode should ideally work for any ethereum network.

If we move network-specific constants and behavior into .yml files, we can trivially support locally run testnets, which will be important for e2e test automation. Additionally, chores like https://github.com/rocket-pool/smartnode/pull/437 become trivial- only one file needs to be updated, with optional "what's changed" update elsewhere.

We can package some 'default' networks with smartnode-install, and read an optional extra-networks.yml from the project directory (~/.rocketpool) 

Edit: additionally, putting everything that fornax had to delete in

activescott:
Submission for milestone B of [bounty BA0902402](https://rpbountyboard.com/BA0902402) to move networks to yaml.

The networks-default.yml file I'll put up a PR for in smartnode-install.

fixes #438

@jshufro Can you please review and make sure this is on the right track. I _think_ this is complete. The things that are on my mind though:
- Not super clear what would be an effective set of tests as much of that data is used in operations that I'm not sure how to easily/quickly exercise. If you want to suggest some test cases I'm happy to run them. I was able to use `rocketpool service config` and restart my holesky node and it seems fine.
-Am I missing some other places where network info is needed?

### DIFF
--- a/rocketpool-cli/rocketpool-cli.go
+++ b/rocketpool-cli/rocketpool-cli.go
@@ -145,6 +145,8 @@ ______           _        _    ______           _
 		if cfg.Smartnode.GetRplFaucetAddress() != "" {
 			faucet.RegisterCommands(app, "faucet", []string{"f"})
 		}
+
+		service.RegisterCommands(app, "service", []string{"s"}, cfg)
 	}
 
 	minipool.RegisterCommands(app, "minipool", []string{"m"})
@@ -152,7 +154,6 @@ ______           _        _    ______           _
 	node.RegisterCommands(app, "node", []string{"n"})
 	odao.RegisterCommands(app, "odao", []string{"o"})
 	queue.RegisterCommands(app, "queue", []string{"q"})
-	service.RegisterCommands(app, "service", []string{"s"})
 	wallet.RegisterCommands(app, "wallet", []string{"w"})
 
 	app.Before = func(c *cli.Context) error {
--- a/shared/services/config/networks-config.go
+++ b/shared/services/config/networks-config.go
@@ -1,0 +1,65 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/rocket-pool/smartnode/shared/types/config"
+	"gopkg.in/yaml.v2"
+)
+
+const defaultNetworksConfigPath = "networks-default.yml"
+
+type NetworksConfig struct {
+	Networks []*config.NetworkInfo `yaml:"networks"`
+}
+
+func NewNetworksConfig() *NetworksConfig {
+	return &NetworksConfig{
+		Networks: make([]*config.NetworkInfo, 0),
+	}
+}
+
+func LoadNetworksFromFile(configPath string) (*NetworksConfig, error) {
+	_, err := os.Stat(configPath)
+	if os.IsNotExist(err) {
+		return nil, fmt.Errorf("config path does not exist: %s", configPath)
+	}
+
+	filePath := path.Join(configPath, defaultNetworksConfigPath)
+	if os.IsNotExist(err) {
+		return nil, fmt.Errorf("default networks file does not exist: %s", filePath)
+	}
+
+	bytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not read default networks file: %w", err)
+	}
+
+	networks := NetworksConfig{}
+	err = yaml.Unmarshal(bytes, &networks)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal default networks file: %w", err)
+	}
+
+	return &networks, nil
+}
+
+func (nc *NetworksConfig) GetNetworks() []*config.NetworkInfo {
+	return nc.Networks
+}
+
+func (nc *NetworksConfig) SaveToFile(filePath string) error {
+	// TODO: this signature doesn't support saving to default/extras files
+	d, err := yaml.Marshal(nc)
+	if err != nil {
+		return fmt.Errorf("could not marshal networks config: %w", err)
+	}
+
+	err = os.WriteFile(filePath, d, 0644)
+	if err != nil {
+		return fmt.Errorf("could not write networks config to file to %s: %w", filePath, err)
+	}
+	return nil
+}
--- a/shared/services/config/rocket-pool-config.go
+++ b/shared/services/config/rocket-pool-config.go
@@ -58,6 +58,8 @@ type RocketPoolConfig struct {
 
 	IsNativeMode bool `yaml:"-"`
 
+	networks *NetworksConfig `yaml:"-"`
+
 	// Execution client settings
 	ExecutionClientMode config.Parameter `yaml:"executionClientMode,omitempty"`
 	ExecutionClient     config.Parameter `yaml:"executionClient,omitempty"`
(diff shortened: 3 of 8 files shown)


rocketpool-cli/rocketpool-cli.go
```diff
@@ -145,14 +145,15 @@ ______           _        _    ______           _
...
+
+		service.RegisterCommands(app, "service", []string{"s"}, cfg)
 	}
 
 	minipool.RegisterCommands(app, "minipool", []string{"m"})
 	network.RegisterCommands(app, "network", []string{"e"})
 	node.RegisterCommands(app, "node", []string{"n"})
 	odao.RegisterCommands(app, "odao", []string{"o"})
 	queue.RegisterCommands(app, "queue", []string{"q"})
-	service.RegisterCommands(app, "service", []string{"s"})
```
jshufro, line 155: this feels like it will break... a lot

jshufro (commented):
Directionally, looks good. Might be neater to move all the `Address` fields into a substruct of NetworkInfo

activescott:
@jshufro Added substruct. I also noticed I forgot to add support for extra-networks.yml and did that and tested it. Please let me know what else needs done here to get this one merged.

shared/services/config/networks-config.go
```diff
@@ -0,0 +1,71 @@
...
+	extraNetworks, err := loadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not load extra networks file: %w", err)
+	}
+	defaultNetworks.Networks = append(defaultNetworks.Networks, extraNetworks.Networks...)
+
+	return defaultNetworks, nil
+}
+
+func loadFile(filePath string) (*NetworksConfig, error) {
```
jshufro, line 54: since this is all in the `config` package we should namespace this, `loadNetworksFile`

shared/services/config/networks-config.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/rocket-pool/smartnode/shared/types/config"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	defaultNetworksConfigPath = "networks-default.yml"
+	extraNetworksConfigPath   = "networks-extra.yml"
+)
+
+type NetworksConfig struct {
```
jshufro, line 17: seems like a great opportunity to use `type NetworksConfig []*config.NetworkInfo`.

You'll have to remove the `networks:` piece from the smartnode-install .yml file and just have the list be top-level.

but then, instead of:

`defaultNetworks.Networks = append(defaultNetworks.Networks, extraNetworks.Networks...)` below you can ergonmically `return append(defaultNetworks, extraNetworks...)`

and you can kill off `func (nc *NetworksConfig) GetNetworks() []*config.NetworkInfo `

shared/services/config/rocket-pool-config.go
```diff
@@ -165,9 +167,13 @@ func LoadFromFile(path string) (*RocketPoolConfig, error) {
 	if err := yaml.Unmarshal(configBytes, &settings); err != nil {
 		return nil, fmt.Errorf("could not parse settings file: %w", err)
 	}
-
+	// now load the networks:
+	networks, err := LoadNetworksFromFile(filepath.Dir(path))
+	if err != nil {
+		return nil, fmt.Errorf("could not load networks file: %w", err)
+	}
```
jshufro, line 174: kinda seems like `NewRocketPoolConfig` should handle this. `filepath.Dir(path)` gets stored in `cfg.RocketPoolDirectory` on alloc, and then in `cfg.Deserialize` you can parse the new yml.

shared/services/config/smartnode-config.go
```diff
@@ -622,16 +336,28 @@ func (cfg *SmartnodeConfig) GetParameters() []*config.Parameter {
 
 // Getters for the non-editable parameters
 
+// Gets the currently chosen network's configuration
+func (cfg *SmartnodeConfig) GetNetworkInfo() *config.NetworkInfo {
+	for _, network := range cfg.parent.networks.GetNetworks() {
+		networkName := fmt.Sprintf("%s", cfg.Network.Value)
```
jshufro, line 342: hmm, when we parse the .yml files maybe we should save them in a `map[config.Network]NetworkInfo` that we can access here without the loop.

shared/services/config/smartnode-config.go
```diff
@@ -804,113 +530,105 @@ func (cfg *SmartnodeConfig) GetFeeRecipientFilePath() string {
...
 }
 
 func (cfg *SmartnodeConfig) GetV110MinipoolFactoryAddress() common.Address {
-	return common.HexToAddress(cfg.v1_1_0_MinipoolFactoryAddress[cfg.Network.Value.(config.Network)])
+	return common.HexToAddress(cfg.GetNetworkInfo().Addresses.V1_1_0_MinipoolFactory)
 }
 
 func (cfg *SmartnodeConfig) GetPreviousRewardsPoolAddresses() []common.Address {
-	return cfg.previousRewardsPoolAddresses[cfg.Network.Value.(config.Network)]
+	addresses := []common.Address{}
```
jshufro, line 569: incredibly premature optimization from me:
```suggestion
	addresses := make([]common.Address, 0, len(cfg.GetNetworkInfo().Addresses.PreviousRewardsPools))
```

shared/services/config/networks-config.go
activescott: My thought here was future proofing: We may want other information about that config file rather than just an array of NetworkInfo at some point. 

I don't feel strongly though. Let me know if you prefer though the flat array and I can change it.

jshufro: Good point, it's easier to extend if the root level is an object. I kind of am leaning towards:
```golang
type NetworksConfigs []*config.NetworkInfo
type NetworksConfigFile struct {
	Networks []NetworkConfigs `yaml:"networks"`
}
```
now... that way we get the extensibility of the file format _and_ can simplify the ABI here to remove GetNetworks().

If we ever have to add more fields in memory and change the `NetworkConfigs` type down the line, the compiler will yell at us about the surface area, so it's any easy upgrade path.

shared/services/config/rocket-pool-config.go
activescott: I've looked into this again and I prefer passing it into New for two reasons... `NewRocketPoolConfig` [calls `NewSmartnodeConfig`](https://github.com/activescott/smartnode/blob/cef7336fc59d22944c1a5de56fbec89b862a18c7/shared/services/config/rocket-pool-config.go#L458) which [loads a list of network options](https://github.com/activescott/smartnode/blob/cef7336fc59d22944c1a5de56fbec89b862a18c7/shared/services/config/smartnode-config.go#L155) so it requires networks to be available already.

There is also a place in [the client creates a fresh config without deserializing](https://github.com/activescott/smartnode/blob/cef7336fc59d22944c1a5de56fbec89b862a18c7/shared/services/rocketpool/client.go#L230) that would require some patching after it is created to even have a valid config. So I think it's more correct to have networks passed in.

Let me know if I'm missing something.

jshufro: Can we meet in the middle and have `NewRocketPoolConfig` parse the files just before calling `NewSmartnodeConfig`?

In both places NewRocketPoolConfig is called it uses the same path as the previous LoadNetworksFromeFile call

activescott: :) I agree with your edit. 

There is a third place where that `NewRocketPoolConfig` that actually doesn't require reading the networks: https://github.com/activescott/smartnode/blob/cef7336fc59d22944c1a5de56fbec89b862a18c7/shared/services/rocketpool/client.go#L230

After reviewing this, I prefer the way it is now. Let me know what you prefer though. I'm open.

shared/services/config/networks-config.go
jshufro: Seems like a good option as well. I think this one is ready for @0xfornax but i'll give it one last pass

jshufro (commented):
Can you do a brief bit of research to see how hard it is to delete:
```
// Enum to describe which network the system is on
const (
        Network_Unknown Network = ""
        Network_All     Network = "all"
        Network_Mainnet Network = "mainnet"
        Network_Devnet  Network = "devnet"
        Network_Holesky Network = "holesky"
)
```

The TUI will have to generate settings according to the parsed network, but beyond that, there isn't a ton of value in moving network-related settings to yml when the list of networks itself remains hardcoded.

`Network_All` of course can stay

activescott:
Hey @jshufro. You mentioned holding off on this. Should I pick up those changes from comments and see this through or is the consensus that we should no longer do this?

jshufro:
I think the best thing would be to rebase onto the `v2` branch. A lot of the work unfortunately won't apply, but the good and bad news is that this was already incomplete- https://github.com/rocket-pool/smartnode/pull/459#pullrequestreview-1935169929 I do think for this to be meaningful we need to be able to add networks simply by adding yaml

PR #466, opened 2024-03-24
Move dev into v2

jclapis:
As per Patches, we'll start treating `v2` like `master` and only put things into it via PRs instead of committing directly.


jshufro (approved):
<3

PR #480, opened 2024-04-08
Fix a few bugs with the migration

jshufro:
`UpdateConfig` needs to actually return the updated config so it can be used by the upstream caller.

`configVersionString` was being clobbered.


jclapis:
Good catches, thanks Patches.

PR #481, opened 2024-04-08
Push users towards pbss

jshufro:
Adds a friendly message suggesting that users should resync geth with PBSS enabled instead of pruning it.

Deletes two redundant color defines

### DIFF
--- a/rocketpool-cli/service/service.go
+++ b/rocketpool-cli/service/service.go
@@ -162,8 +162,6 @@ func installUpdateTracker(c *cli.Context) error {
 	}
 
 	// Print success message & return
-	colorReset := "\033[0m"
-	colorYellow := "\033[33m"
 	fmt.Println("")
 	fmt.Println("The Rocket Pool update tracker service was successfully installed!")
 	fmt.Println("")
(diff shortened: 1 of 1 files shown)


rocketpool-cli/service/service.go
```diff
@@ -991,9 +989,12 @@ func pruneExecutionClient(c *cli.Context) error {
 	}
 
 	if selectedEc == cfgtypes.ExecutionClient_Geth || selectedEc == cfgtypes.ExecutionClient_Besu {
+		if selectedEc == cfgtypes.ExecutionClient_Geth {
+			fmt.Printf("%sGeth has a new feature that renders pruning obsolete. Consider enabling PBSS in the Execution Client settings in `rocketpool service config` and resyncing with `rocketpool service resync-eth1` instead of pruning.%s", colorYellow, colorReset)
```
jshufro, line 993:
```suggestion
			fmt.Printf("%sGeth has a new feature that renders pruning obsolete. Consider enabling PBSS in the Execution Client settings in `rocketpool service config` and resyncing with `rocketpool service resync-eth1` instead of pruning.%s\n", colorYellow, colorReset)
```

PR #496, opened 2024-04-17
Add goimports lint github action

jshufro:
Annotates the un-formatted files:
image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,50 @@
+# Taken from https://github.com/golangci/golangci-lint-action
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,50 @@
...
+          version: latest
+
+          # Optional: working directory, useful for monorepos
+          # working-directory: somedir
+
+          # Optional: golangci-lint command line arguments.
+          # args: --issues-exit-code=0
+
+          # For now, Smart Node will only enforce goimports linting
+          args: --disable-all --enable goimports
```
jshufro, line 37: This is the only line that differs from the sample config- for now, let's not get slap-happy with static analyses. Once v2 is landed we can crank them up.

PR #498, opened 2024-04-17
Port lint and test CI to v2

jshufro:


rocketpool-daemon/watchtower/submit-rpl-price.go
```diff
@@ -82,7 +82,7 @@ func (t *SubmitRplPrice) Run(state *state.NetworkState) error {
 	err := t.updateL2Prices(state)
 	if err != nil {
 		// Error is not fatal for this task so print and continue
-		t.logger.Error("Error updating L2 prices: %s", err.Error())
+		t.logger.Error("Error updating L2 prices", "error", err.Error())
```
jclapis, line 85: For consistency, the other times errors or logged use `log.Err(err)` which has the `error` key built in rather than specifying it explicitly.

jclapis (approved):

PR #500, opened 2024-04-17
Enable build ci. Enable staticcheck, ineffassign, gosimple, unused linters and fix their errors

jshufro:
The very last commit should be reviewed with care: https://github.com/rocket-pool/smartnode/pull/500/commits/c235da8ca8ae73a1f5cc57777d00ef440c821e37

When deleting unused code, it's not obvious that it is _intentionally_ unused.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,8 +34,8 @@ jobs:
           # Optional: golangci-lint command line arguments.
           # args: --issues-exit-code=0
 
-          # For now, Smart Node will only enforce goimports linting
-          args: --disable-all --enable goimports --enable staticcheck --enable ineffassign --enable gosimple --enable unused
+          # For now, Smart Node will enforce everything except errcheck
+          args: --disable errcheck
 
           # Optional: show only new issues if it's a pull request. The default value is `false`.
           # only-new-issues: true


rocketpool-daemon/common/rewards/prater-events.go
```diff
@@ -1,132 +0,0 @@
-package rewards
```
jclapis, line 1: o7

rocketpool-daemon/common/state/network-state.go
```diff
@@ -337,7 +337,8 @@ func CreateNetworkStateForNode(cfg *config.SmartNodeConfig, rp *rocketpool.Rocke
 		return nil, nil, fmt.Errorf("error getting Protocol DAO proposal details: %w", err)
 	}
 	logger.Info(fmt.Sprintf("%d/%d - Retrieved Protocol DAO proposals", currentStep, steps), slog.Duration(keys.TotalElapsedKey, time.Since(start)))
-	currentStep++
+	// If adding more steps below, don't forget to increment currentStep:
```
jclapis, line 340: Does this feel like a footgun? The `currentStep++` at the end wasn't hurting anything and prevents someone from messing it up in the future if they add more steps.

jshufro: ineffassign doesn't like nonfunctional value assignments to variables.

It is a footgun, because the abstraction is poor- the maintainer shouldn't need to remember to increment something when moving between steps.

Instead, they should formally define the steps in a slice, and they should execute in a `for` loop. That is, however, out of scope for this change. I'll file to follow-up.

rocketpool-cli/commands/service/config/step-random-bn.go
```diff
@@ -9,43 +9,6 @@ import (
 const randomBnPrysmID string = "step-random-bn-prysm"
 const randomBnID string = "step-random-bn"
 
-func createRandomPrysmStep(wiz *wizard, currentStep int, totalSteps int, goodOptions []*config.ParameterOption[config.BeaconNode]) *choiceWizardStep {
```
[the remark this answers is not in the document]
jclapis: Okey dokey, just have to remember when to roll back to if we ever need it again.

rocketpool-daemon/api/pdao/utils.go
```diff
@@ -32,25 +32,3 @@ func createPollard(context context.Context, logger *slog.Logger, rp *rocketpool.
 	}
 	return blockNumber, pollard, nil
 }
-
-// Loads (or regenerates) the pollard for a proposal from a block number
```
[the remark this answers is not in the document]
jclapis: Ok, yeah the disk cache got rolled into `createPollard` so this can go.

rocketpool-daemon/common/state/network-state.go
jclapis: Ok, let's take it out for now and I'll await the follow-up refactor.

jclapis (approved):

PR #505, opened 2024-04-21
Make alerting package stateful and directly error-logging. Fix the rest of the unchecked errors.

jshufro:
The `alerting` package previously was stateless and simply recreated state every time any alert was called.

Now it is stateful, and only checks config on init(). Logging is done through a logger instead of through stdout.

Fixes `errcheck` issues and uses newest `rocketpool-go/v2` and `node-manager-core` commits which resolve upstream errcheck issues.

Enables all linters :tada:

### DIFF
--- a/rocketpool-cli/commands/service/config.go
+++ b/rocketpool-cli/commands/service/config.go
@@ -131,8 +131,13 @@ func configureService(c *cli.Context) error {
 				suffix := config.GetContainerName(container)
 				name := cfg.GetDockerArtifactName(suffix)
 				fmt.Printf("Stopping %s... ", name)
-				rp.StopContainer(name)
-				fmt.Print("done!\n")
+				err := rp.StopContainer(name)
+				if err != nil {
+					fmt.Println("error!")
+					fmt.Fprintf(os.Stderr, "Error stopping container %s: %s\n", name, err.Error())
+					continue
+				}
+				fmt.Println("done!")
 			}
 
 			fmt.Println()
--- a/rocketpool-daemon/node/metrics-exporter.go
+++ b/rocketpool-daemon/node/metrics-exporter.go
@@ -71,7 +71,7 @@ func runMetricsServer(ctx context.Context, sp *services.ServiceProvider, logger 
 	metricsPath := "/metrics"
 	http.Handle(metricsPath, handler)
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		w.Write([]byte(`<html>
+		_, err := w.Write([]byte(`<html>
             <head><title>Rocket Pool Metrics Exporter</title></head>
             <body>
             <h1>Rocket Pool Metrics Exporter</h1>
@@ -79,6 +79,10 @@ func runMetricsServer(ctx context.Context, sp *services.ServiceProvider, logger 
             </body>
             </html>`,
 		))
+
+		if err != nil {
+			logger.Warn("Error replying to http request in metrics exporter", log.Err(err))
+		}
 	})
 
 	// Run the server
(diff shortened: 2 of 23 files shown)


jclapis (approved):
LGTM, have a few error messages to embellish but they won't block the PR; they can come later with the master merge.

rocketpool-cli/commands/service/config.go
```diff
@@ -131,8 +131,14 @@ func configureService(c *cli.Context) error {
 				suffix := config.GetContainerName(container)
 				name := cfg.GetDockerArtifactName(suffix)
 				fmt.Printf("Stopping %s... ", name)
-				rp.StopContainer(name)
-				fmt.Print("done!\n")
+				err := rp.StopContainer(name)
+				if err != nil {
+					fmt.Print("error!")
+					fmt.Fprintf(os.Stderr, "Error stopping container %s: %s\n", name, err.Error())
```
jclapis, line 137: Is this going to end up printing to the terminal on the same line as the above since there's not a newline first?

rocketpool-daemon/node/metrics-exporter.go
```diff
@@ -71,14 +71,18 @@ func runMetricsServer(ctx context.Context, sp *services.ServiceProvider, logger
...
             <head><title>Rocket Pool Metrics Exporter</title></head>
             <body>
             <h1>Rocket Pool Metrics Exporter</h1>
             <p><a href='` + metricsPath + `'>Metrics</a></p>
             </body>
             </html>`,
 		))
+
+		if err != nil {
+			logger.Warn("Error replying to http request in metrics exporter", log.Err(err))
```
jclapis, line 84: Let's add some more context here, such as where the request came from and its path (prob not the entire thing because the query string can get really big).

rocketpool-daemon/node/node.go
```diff
@@ -171,13 +173,15 @@ func (t *TaskLoop) Run() error {
...
 
 // Stop the daemon
-func (t *TaskLoop) Stop() {
+func (t *TaskLoop) Stop() error {
 	if t.metricsServer != nil {
 		// Shut down the metrics server
 		ctx, cancel := context.WithTimeout(context.Background(), metricsShutdownTimeout)
 		defer cancel()
-		t.metricsServer.Shutdown(ctx)
+		return t.metricsServer.Shutdown(ctx)
```
jclapis, line 181: This was "intentionally" left this way because we're shutting the server down anyway so if the HTTP server errors out during shutdown there's nothing to be done about it but it doesn't hurt to log it. Let's add a wrapper noting the error happened during HTTP server shutdown here.

rocketpool-daemon/node/node.go
```diff
@@ -317,7 +321,12 @@ func (t *TaskLoop) runTasks() bool {
 
 	// Run watchtower duties in parallel
 	var watchtowerWg errgroup.Group
-	defer watchtowerWg.Wait()
+	defer func() {
+		err := watchtowerWg.Wait()
+		if err != nil {
+			t.logger.Error(err.Error())
```
jclapis, line 327: This could use a wrapper noting the error happened in the watchtower loop.

jclapis (commented):
Added said embellishments in as comments.

rocketpool-cli/commands/service/config.go
jshufro: if stdout and stderr are the same file, yeah. we'll have to hoist the err message outside the loop, or move the blank println back in

rocketpool-daemon/node/metrics-exporter.go
jshufro: the path is `/`, on L73 we specifically only handle `/` with this body

jclapis (approved):
Cheers!

PR #506, opened 2024-04-21
Added "is RPL Locking Allowed" check to every `PDAOCanProposeXYZ` function 

thomaspanf:
PDAO functions edited:

- Propose Setting
- Propose Sec Council Invite
- Propose Sec Council Kick 
- Propose One Time Spend
- Propose Rewards Percentages
- Propose Recurring Spend
- Propose Recurring Spend Update
- Propose Sec Council Update

~~Haven't tested these 3 yet:~~

~~- Propose Sec Council Kick~~
~~- Propose Recurring Spend Update~~
~~- Propose Sec Council Update~~

fixed usage of goroutines per Fornax's suggestion!

### DIFF
--- a/rocketpool/api/pdao/replace-security.go
+++ b/rocketpool/api/pdao/replace-security.go
@@ -6,10 +6,12 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/rocket-pool/rocketpool-go/dao/protocol"
 	"github.com/rocket-pool/rocketpool-go/dao/security"
+	"github.com/rocket-pool/rocketpool-go/node"
 	"github.com/rocket-pool/smartnode/shared/services"
 	"github.com/rocket-pool/smartnode/shared/types/api"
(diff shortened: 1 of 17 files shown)


rocketpool/api/pdao/replace-security.go
```diff
@@ -35,6 +37,37 @@ func canProposeReplaceMemberOfSecurityCouncil(c *cli.Context, existingMemberAddr
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Sync
+	var isRplLockingAllowed bool
+	var wg errgroup.Group
+
+	// Get is RPL locking allowed
+	wg.Go(func() error {
```
0xfornax, line 50: You usually want to use goroutines when starting multiple queries at the same time. If you have only one query then you don't need to use threads.

thomaspanf: Thanks for the tip! I've fixed the usage of goroutines.

0xfornax (approved):

PR #532, opened 2024-05-19
Moving away from the deprecated network id and checking chain ids

0xfornax:


jshufro:
Why is checkChainIds passed when it's always 'true'?

shared/services/ec-manager.go
```diff
@@ -35,8 +35,10 @@ type ExecutionClientManager struct {
 type ecFunction func(*ethclient.Client) (interface{}, error)
 
 // Creates a new ExecutionClientManager instance based on the Rocket Pool config
-func NewExecutionClientManager(cfg *config.RocketPoolConfig) (*ExecutionClientManager, error) {
-
+func NewExecutionClientManager(cfg *config.RocketPoolConfig, checkChainIDs bool) (*ExecutionClientManager, error) {
```
jshufro, line 38: `checkChainIDs` still appears here and is always `true`

jshufro (approved):

PR #539, opened 2024-05-31
chore: fix some function names

linghuying:


jshufro:
unfortunately these are generated files

linghuying:
> unfortunately these are generated files

Thanks for your replay.

PR #540, opened 2024-06-01
Fix minipool close crash

0xfornax:


rocketpool-daemon/api/minipool/close-details.go
```diff
@@ -97,10 +97,12 @@ func (c *minipoolCloseDetailsContext) PrepareData(addresses []common.Address, mp
 
 	// Get the node shares
 	err = c.rp.BatchQuery(len(addresses), minipoolCompleteShareBatchSize, func(mc *batch.MultiCaller, i int) error {
-		mpv3, success := minipool.GetMinipoolAsV3(mps[i])
-		if success {
-			details[i].Distributed = mpv3.HasUserDistributed.Get()
-			mpv3.CalculateNodeShare(mc, &details[i].NodeShareOfEffectiveBalance, details[i].EffectiveBalance)
+		if details[i].CanClose {
```
jshufro, line 100:
```suggestion
		if !details[i].CanClose {
		  return nil
		}
```

we should strive to keep indentation manageable

PR #551, opened 2024-06-14
Fix insufficentBalance comparison

pumpchaser:
Currently, InsufficientBalance check will always return false as long as there's a balance in the wallet. This PR updates the logic to compare the wallet balance against the transfer amount.


0xfornax (approved):
lgtm
Thanks for the contribution!

PR #553, opened 2024-06-17
Remove lh deprecated tag

0xfornax:
Removing the `-modern` flag since it was deprecated.


shared/services/config/lighthouse-config.go
```diff
@@ -136,21 +133,3 @@ func (cfg *LighthouseConfig) GetName() string {
...
-	missingFeatures := sys.GetMissingModernCpuFeatures()
-	if len(missingFeatures) > 0 {
-		return lighthouseTagPortableProd
-	}
-	return lighthouseTagModernProd
-}
-
-// Get the appropriate LH default tag for testnets
-func getLighthouseTagTest() string {
-	missingFeatures := sys.GetMissingModernCpuFeatures()
```
jshufro, line 151: Let's remove `GetMissingModernCpuFeatures()` as well. It's only used here, and to complain when someone sets -modern but isn't. (`rocketpool-cli/service/service.go:checkCpuFeatures`)

jshufro (changes requested):

jshufro (approved):

PR #555, opened 2024-06-18
Removed isHoustonDeployed checks.

0xfornax:


rocketpool/api/node/status.go
```diff
@@ -93,7 +86,7 @@ func getStatus(c *cli.Context) (*api.NodeStatusResponse, error) {
 
 	// Get node details
 	wg.Go(func() error {
-		details, err := node.GetNodeDetails(rp, nodeAccount.Address, response.IsHoustonDeployed, nil)
+		details, err := node.GetNodeDetails(rp, nodeAccount.Address, true, nil)
```
jshufro, line 89: I'd prefer to see this argument deleted in rocketpool-go first, and then this PR should have a single commit that updates go.mod/sum and removes calls that include this arg

jshufro:
Odd- I can't find v1.8.3 of rocketpool-go (referenced in go.mod) https://github.com/rocket-pool/rocketpool-go/tags

rocketpool/api/node/status.go
0xfornax: https://github.com/rocket-pool/rocketpool-go/pull/26

0xfornax:
> Odd- I can't find v1.8.3 of rocketpool-go (referenced in go.mod) https://github.com/rocket-pool/rocketpool-go/tags

I've pushed the v1.8.3 tag.

rocketpool/api/network/stats.go
```diff
@@ -128,7 +128,7 @@ func getStats(c *cli.Context) (*api.NetworkStatsResponse, error) {
 	wg.Go(func() error {
 		multicallerAddress := common.HexToAddress(cfg.Smartnode.GetMulticallAddress())
 		balanceBatcherAddress := common.HexToAddress(cfg.Smartnode.GetBalanceBatcherAddress())
-		contracts, err := rpstate.NewNetworkContracts(rp, multicallerAddress, balanceBatcherAddress, false, nil)
```
jshufro, line 131: cool- great example of why it's good to update the library to remove the argument first. we missed this call earlier.

jshufro (approved):
If we wanted to be _really_ pedantic about this, I would insist on using `git rebase` to reorder the commits so the rocketpool-go dependency is updated before the other changes, but meh, up to you.

PR #557, opened 2024-06-19
Handle not in rp nodeset bug

thomaspanf:
- Fix "address is not in the RP node set" (when voting isn't initialized and you call `rp p s`). This causes the call to return instead of printing the rest of the menu.
```
ubuntu@v1stable:~$ rocketpool p s

error after requesting get-voting-power: address 0xa6073aEecB94FB769bB1b3754A0ef8b3bec36199 is not in the RP node set
```
- Add wallet not initialized check
- Add display for selected network

before:
```
tpan@tpan-meerkat:~$ rocketpool p s

error after requesting get-voting-power: Wallet is not initialized
```
after:
```
tpan@tpan-meerkat:~$ rocketpool p s

Your Smart Node is currently using the Holesky Test Network.

The node wallet is not initialized.
```

Also included some small changes to the `pdao status` menu output (e.g missing newlines)

### DIFF
--- a/rocketpool-cli/pdao/status.go
+++ b/rocketpool-cli/pdao/status.go
@@ -78,10 +78,9 @@ func getStatus(c *cli.Context) error {
 	// Onchain Voting Status
 	fmt.Printf("%s=== Onchain Voting ===%s\n", colorGreen, colorReset)
 	if response.IsVotingInitialized {
-		fmt.Printf("The node has %s%s%s been initialized for onchain voting.", colorBlue, response.AccountAddressFormatted, colorReset)
-
+		fmt.Printf("The node has %s%s%s been initialized for onchain voting.\n", colorBlue, response.AccountAddressFormatted, colorReset)
 	} else {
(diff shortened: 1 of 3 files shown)


rocketpool-cli/pdao/status.go
```diff
@@ -93,10 +117,13 @@ func getStatus(c *cli.Context) error {
 	}
 	fmt.Printf("The node's local voting power: %.10f\n", eth.WeiToEth(response.VotingPower))
 
-	fmt.Printf("Total voting power delegated to the node: %.10f\n", eth.WeiToEth(response.TotalDelegatedVp))
+	if response.IsVotingInitialized != false {
+		fmt.Printf("Total voting power delegated to the node: %.10f\n", eth.WeiToEth(response.TotalDelegatedVp))
+	} else {
+		fmt.Print("The node must initialize onchain voting to be eligible to recieve delegated voting power.\n")
```
0xfornax, line 123: A node must only be registered to be able to receive delegated vp.

0xfornax (approved):

PR #559, opened 2024-06-20
Refactor initialize-voting to initialize-voting-with-delegate 

thomaspanf:
```
USAGE:
   rocketpool pdao command [command options] [arguments...]

COMMANDS:
   initialize-voting, iv                 Unlocks a node operator's voting power (only required for node operators who registered before governance structure was in place)
   initialize-voting-with-delegate, ivd  Unlocks a node operator's voting power and sets the delegate (saves some gas by initializing and setting the delegate in one transaction)
```
Added two seperate commands (one for initializing voting, another for initializing voting and setting the delegate in one transaction).

### DIFF
--- a/rocketpool-cli/pdao/commands.go
+++ b/rocketpool-cli/pdao/commands.go
@@ -67,10 +67,22 @@ func RegisterCommands(app *cli.App, name string, aliases []string) {
 				},
 			},
 			{
+				Name:      "initialize-voting",
+				Aliases:   []string{"iv"},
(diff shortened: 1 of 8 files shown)


0xfornax:
I'd rather see it as one command where we ask if they want to use their node address or delegate to someone else.

thomaspanf:
> I'd rather see it as one command where we ask if they want to use their node address or delegate to someone else.

Yeah that's what I initially thought too, two commands was Langer's request. We can bring it up in standup.

0xfornax (approved):

PR #562, opened 2024-06-21
Remove hardcoded withdrawal limit.

thomaspanf:


rocketpool-cli/node/status.go
```diff
@@ -310,7 +310,7 @@ func getStatus(c *cli.Context) error {
 				"RPIP-30 is in effect and the node will gradually earn rewards in amounts above the previous limit of 150%% of bonded ETH. Read more at https://github.com/rocket-pool/RPIPs/blob/main/RPIPs/RPIP-30.md\n")
 			if rplTotalStake > rplWithdrawalLimit {
 				fmt.Printf(
-					"You can now withdraw down to %.6f RPL (60%% of bonded ETH).\n", math.RoundDown(eth.WeiToEth(status.MaximumRplStake), 6))
+					"You can now withdraw down to %.6f RPL.\n", math.RoundDown(eth.WeiToEth(status.MaximumRplStake), 6))
```
0xfornax, line 313: Instead of removing, we should fetch the percentage from the contracts. 
`getMaximumPerMinipoolStake()`

PR #563, opened 2024-06-21
Check if node has initialized vp before closing mps (to avoid nodeETHMatched issues)

0xfornax:


rocketpool-cli/minipool/close.go
```diff
@@ -44,6 +44,7 @@ func closeMinipools(c *cli.Context) error {
 		return nil
 	}
 
+	hasOnly16ETHMps := true
```
jshufro, line 47: we should probably print a message and exit before this if `len(details.Details) == 0`

rocketpool-cli/minipool/close.go
```diff
@@ -72,6 +73,14 @@ func closeMinipools(c *cli.Context) error {
 				unwithdrawnMinipools = append(unwithdrawnMinipools, mp)
 			}
 		}
+		if eth.WeiToEth(mp.DepositBalance) != 16 {
```
jshufro, line 76: we should move this to L61 so we can refactor the `if mp.CanClose` case to contain a `continue`...

0xfornax: I don't see how this would change an specific minipool CanClose info.

jshufro: it would just make the code more future proof

PR #578, opened 2024-07-01
Allow submissions after consensus is reached.

0xfornax:
At some point the Smart Node stopped doing submissions after the prices/balances consensus was reached. This PR restores the desired behavior, allowing oDAO nodes to make submissions after consensus is reached, as a way to monitor their liveness.

### DIFF
--- a/rocketpool/watchtower/submit-network-balances.go
+++ b/rocketpool/watchtower/submit-network-balances.go
@@ -1,7 +1,6 @@
 package watchtower
 
 import (
-	"context"
 	"fmt"
 	"math/big"
 	"strings"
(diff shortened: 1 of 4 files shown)


rocketpool/watchtower/submit-network-balances.go
```diff
@@ -145,9 +145,31 @@ func (t *submitNetworkBalances) run(state *state.NetworkState) error {
...
+		submissionTimestamp = int64(submissionBlock.Time)
+	} else {
+		// Calculate the next submission timestamp
+		submissionTimestamp, err = utils.FindNextSubmissionTimestamp(latestBlockTimestamp, referenceTimestamp, submissionIntervalInSeconds)
+		if err != nil {
+			return err
+		}
+	}
+
+	if targetBlockNumber < lastSubmissionBlock {
```
jshufro, line 167: targetBlockNumber is 0-value on L148 and isn't updated, so it's always 0 here, and this is always 'true'

jshufro (changes requested):
Seems like a good time to break out this logic into a function.

PR #582, opened 2024-07-03
Add CONTRIBUTING.md (Discussion around contribution guidelines)

jshufro:
Here is my proposal for how we should proceed with formal contribution guidelines. Not much has changed from the status quo, but for the explicit rules around when a PR review ought to be required.

We also name the team roles explicitly, and provide high level guidelines to contributors.


jshufro:
I don't have permission to request reviews soooo
@0xfornax @darrenlangley @thomaspanf please discuss.

.github/workflows/commits.yml
jshufro, line 1: added because of the guideline about fixups

PR #595, opened 2024-07-14
Refactor rocketpool-cli.go to improve code organization, make flag parsing testing easier.

jshufro:
- Rename SmartNodeContext to SmartNodeSettings
  - This name seems to be chose because the struct is embedded in cli.Context. However, unlike cli.Context, SmartNodeContext does not extend context.Context, and should not be named as such.
- Moves global cli flags into utils/context
- Renames utils/context to settings/
- Refactors main() to make it more modular
- Adds an assets/ package with text assets (logo, version)
- Adds a test to assets/

WIP: adding tests to rocketpool-cli/ to validate argument parsing logic. Just wanted to get a coverage report back in the meantime.


0xfornax (commented):
A suggestion would be to treat PrintPatchNotes() in a similar way and fetch the "What's new in this version text" from an asset file.

0xfornax (approved):

jshufro:
Rebased (fixups squashed)

PR #599, opened 2024-07-17
Implement pdao status and signalling address commands 

thomaspanf:
`rocketpool pdao status`
`rocketpool pdao set-signalling-address`
`rocketpool pdao clear-signalling-address`

### DIFF
--- a/rocketpool-cli/commands/pdao/signalling-address.go
+++ b/rocketpool-cli/commands/pdao/signalling-address.go
@@ -1,7 +1,13 @@
 package pdao
 
 import (
+	"encoding/hex"
+	"fmt"
+
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/rocket-pool/smartnode/v2/rocketpool-cli/client"
+	"github.com/rocket-pool/smartnode/v2/rocketpool-cli/utils/tx"
+	"github.com/rocket-pool/smartnode/v2/shared/utils"
 	"github.com/urfave/cli/v2"
 )
 
@@ -15,20 +21,54 @@ func setSignallingAddress(c *cli.Context, signallingAddress common.Address, sign
 	// // Build the TX
 	// response, err := rp.Api.PDao.SetSignallingAddress()
 
+	// Test Strings
+	fmt.Printf("Signalling Address: %s\n", signallingAddress)
+	fmt.Printf("Signature: %s\n", signature)
+	fmt.Println()
+
+	sig, err := utils.ParseEIP712(signature)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("EIP712Components:")
+	fmt.Printf("V: %d\n", sig.V)
+	fmt.Printf("R: %s\n", hex.EncodeToString(sig.R[:]))
+	fmt.Printf("S: %s\n", hex.EncodeToString(sig.S[:]))
+
 	return nil
 
 }
 
 func clearSignallingAddress(c *cli.Context) error {
 
-	// // Get RP client
-	// rp, err := client.NewClientFromCtx(c)
-	// if err != nil {
-	// 	return err
-	// }
+	// Get RP client
+	rp, err := client.NewClientFromCtx(c)
+	if err != nil {
+		return err
+	}
 
-	// // Build the TX
-	// response, err := rp.Api.PDao.ClearSignallingAddress()
+	// Build the TX
+	response, err := rp.Api.PDao.ClearSignallingAddress()
 
+	// Test Strings
+	fmt.Printf("Test\n")
+	fmt.Printf("Response %v\n", response) // response returns nil
+	fmt.Printf("Txinfo %v\n", response.Data.TxInfo)
+
+	validated, err := tx.HandleTx(c, rp, response.Data.TxInfo,
+		"Are you sure you want to clear your current snapshot address?",
+		"clearing snapshot address",
+		"Clearing snapshot address...",
+	)
+	if err != nil {
+		return err
+	}
+	if !validated {
+		return nil
+	}
+
+	// Log & return
+	fmt.Println("The node's signalling address has been sucessfully cleared.")
 	return nil
 }
--- a/rocketpool-cli/commands/pdao/signalling-address.go
+++ b/rocketpool-cli/commands/pdao/signalling-address.go
@@ -1,28 +1,34 @@
 package pdao
 
 import (
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/urfave/cli/v2"
 )
 
-func setSignallingAddress(c *cli.Context, signallingAddressString string, signature string) error {
+func setSignallingAddress(c *cli.Context, signallingAddress common.Address, signature string) error {
+	// // Get RP client
+	// rp, err := client.NewClientFromCtx(c)
+	// if err != nil {
+	// 	return err
+	// }
 
-	// // Input Validation
-	// var err error
-	// signallingAddress, err := input.ValidateAddress("signalling-address", signallingAddressString)
-	// if err != nil {
-	// 	return err
-	// }
-	// // Todo: input.ValidateSignature
-	// sig, err := input.ValidateSignature("signature", signature)
-	// if err != nil {
-	// 	return err
-	// }
+	// // Build the TX
+	// response, err := rp.Api.PDao.SetSignallingAddress()
 
 	return nil
 
 }
 
 func clearSignallingAddress(c *cli.Context) error {
+
+	// // Get RP client
+	// rp, err := client.NewClientFromCtx(c)
+	// if err != nil {
+	// 	return err
+	// }
+
+	// // Build the TX
+	// response, err := rp.Api.PDao.ClearSignallingAddress()
 
 	return nil
 }
--- a/rocketpool-cli/commands/pdao/signalling-address.go
+++ b/rocketpool-cli/commands/pdao/signalling-address.go
@@ -1,40 +1,38 @@
 package pdao
 
 import (
-	"encoding/hex"
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/rocket-pool/smartnode/v2/rocketpool-cli/client"
 	"github.com/rocket-pool/smartnode/v2/rocketpool-cli/utils/tx"
-	"github.com/rocket-pool/smartnode/v2/shared/utils"
 	"github.com/urfave/cli/v2"
 )
 
 func setSignallingAddress(c *cli.Context, signallingAddress common.Address, signature string) error {
-	// // Get RP client
-	// rp, err := client.NewClientFromCtx(c)
-	// if err != nil {
-	// 	return err
-	// }
-
-	// // Build the TX
-	// response, err := rp.Api.PDao.SetSignallingAddress()
-
-	// Test Strings
-	fmt.Printf("Signalling Address: %s\n", signallingAddress)
-	fmt.Printf("Signature: %s\n", signature)
-	fmt.Println()
-
-	sig, err := utils.ParseEIP712(signature)
+	// Get RP client
+	rp, err := client.NewClientFromCtx(c)
 	if err != nil {
 		return err
 	}
 
-	fmt.Println("EIP712Components:")
-	fmt.Printf("V: %d\n", sig.V)
-	fmt.Printf("R: %s\n", hex.EncodeToString(sig.R[:]))
-	fmt.Printf("S: %s\n", hex.EncodeToString(sig.S[:]))
+	// Build the TX
+	response, err := rp.Api.PDao.SetSignallingAddress(signallingAddress, signature)
+	if err != nil {
+		return fmt.Errorf("Error setting the signalling address: %w", err)
+	}
+
+	validated, err := tx.HandleTx(c, rp, response.Data.TxInfo,
+		"Are you sure you want to set your signalling address?",
+		"setting signalling",
+		"Setting signalling address...",
+	)
+	if err != nil {
+		return err
+	}
+	if !validated {
+		return nil
+	}
 
 	return nil
 
@@ -50,16 +48,14 @@ func clearSignallingAddress(c *cli.Context) error {
 
 	// Build the TX
 	response, err := rp.Api.PDao.ClearSignallingAddress()
-
-	// Test Strings
-	fmt.Printf("Test\n")
-	fmt.Printf("Response %v\n", response) // response returns nil
-	fmt.Printf("Txinfo %v\n", response.Data.TxInfo)
+	if err != nil {
+		return fmt.Errorf("Error clearing the signalling address: %w", err)
+	}
 
 	validated, err := tx.HandleTx(c, rp, response.Data.TxInfo,
-		"Are you sure you want to clear your current snapshot address?",
-		"clearing snapshot address",
-		"Clearing snapshot address...",
+		"Are you sure you want to clear the current signalling address?",
+		"clearing signalling address",
+		"Clearing signalling address...",
 	)
 	if err != nil {
 		return err
--- a/rocketpool-cli/commands/pdao/signalling-address.go
+++ b/rocketpool-cli/commands/pdao/signalling-address.go
@@ -1,0 +1,28 @@
+package pdao
+
+import (
+	"github.com/urfave/cli/v2"
+)
+
+func setSignallingAddress(c *cli.Context, signallingAddressString string, signature string) error {
+
+	// // Input Validation
+	// var err error
+	// signallingAddress, err := input.ValidateAddress("signalling-address", signallingAddressString)
+	// if err != nil {
+	// 	return err
+	// }
+	// // Todo: input.ValidateSignature
+	// sig, err := input.ValidateSignature("signature", signature)
+	// if err != nil {
+	// 	return err
+	// }
+
+	return nil
+
+}
+
+func clearSignallingAddress(c *cli.Context) error {
+
+	return nil
+}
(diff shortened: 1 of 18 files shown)


rocketpool-daemon/common/contracts/rocket-signer-registry.go
```diff
@@ -38,11 +38,11 @@ type RocketSignerRegistry struct {
 func NewRocketSignerRegistry(address common.Address, client eth.IExecutionClient, txMgr *eth.TransactionManager) (*RocketSignerRegistry, error) {
 	// Parse the ABI
 	var err error
-	snapshotOnce.Do(func() {
+	rocketSignerRegistryOnce.Do(func() {
 		var parsedAbi abi.ABI
 		parsedAbi, err = abi.JSON(strings.NewReader(rocketSignerRegistryAbiString))
 		if err == nil {
-			snapshotAbi = parsedAbi
```
jshufro: and this is why we don't copy paste :laughing:

rocketpool-cli/commands/pdao/signalling-address.go
```diff
@@ -15,20 +21,54 @@ func setSignallingAddress(c *cli.Context, signallingAddress common.Address, sign
...
+	fmt.Printf("Signalling Address: %s\n", signallingAddress)
+	fmt.Printf("Signature: %s\n", signature)
+	fmt.Println()
+
+	sig, err := utils.ParseEIP712(signature)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("EIP712Components:")
```
jshufro: This should _absolutely_ by a receiver on sig.

That is, write a:
`func (e *EIP712Components) String() string {` function to basically just do
`return fmt.Sprintf("%+v", e)`

But, critically, the logic that pretty-prints the struct does not belong in signalling-address.go, it belongs along side the definition of EIP712Components.

rocketpool-daemon/api/pdao/clear-signalling-address.go
```diff
@@ -38,15 +40,150 @@ func (f *protocolDaoClearSignallingAddressFactory) RegisterRoute(router *mux.Rou
...
+	}, nil)
+	if err != nil {
+		return types.ResponseStatus_Error, fmt.Errorf("error getting the registry contract: %w", err)
+	}
 
-func (c *protocolDaoClearSignallingAddressContext) PrepareData(data *api.ProtocolDAOClearSignallingAddressResponse, opts *bind.TransactOpts) (types.ResponseStatus, error) {
+	// Return if there if no signalling address is set
+	if c.signallingAddress == (common.Address{}) {
+		return types.ResponseStatus_Error, nil
+	} else {
```
jshufro, line 77: no else after return

jshufro (changes requested):
PR is still in draft mode so I can't be too perturbed by the presence of commented out code, but good lord that's a lot of commented out code :)

rocketpool-daemon/api/pdao/status.go
```diff
@@ -107,28 +117,32 @@ func (c *protocolDaoGetStatusContext) GetState(mc *batch.MultiCaller) {
 		c.node.Exists,
 		c.node.IsVotingInitialized,
 	)
+
+	// Snapshot Registry
+	if c.registry != nil {
```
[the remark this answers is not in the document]
thomaspanf: I ended up removing this check because it's already done here: https://github.com/rocket-pool/smartnode/blob/8a91b72353d7d647fa02202a1e5908f4a57b634f/rocketpool-daemon/api/pdao/status.go#L84

rocketpool-cli/commands/pdao/status.go
```diff
@@ -12,13 +12,15 @@ import (
...
 )
 
 const (
-	colorBlue             string = "\033[36m"
-	colorReset            string = "\033[0m"
-	colorGreen            string = "\033[32m"
+	colorBlue  string = terminal.ColorBlue
+	colorReset string = terminal.ColorReset
+	colorGreen string = terminal.ColorGreen
+
```
jshufro, line 24: heh. I meant get rid of these consts. Change references to them to refer to the consts in `terminal`.

rocketpool-cli/commands/pdao/status.go
```diff
@@ -32,6 +33,29 @@ func getStatus(c *cli.Context) error {
...
+	walletStatus := statusResponse.Data.WalletStatus
+
+	// Print what network we're on
+	err = utils.PrintNetwork(cfg.Network.Value, isNew)
+	if err != nil {
+		return err
+	}
+	// rp.rp.Api.PDao.GetStatus() will fail with an error, but we can short-circuit it here.
+	if !walletStatus.Address.HasAddress {
+		return errors.New("No node address is loaded.")
```
jshufro, line 56: Perhaps "Node Wallet is not initialized" will be clearer

rocketpool-daemon/api/pdao/status.go
```diff
@@ -125,13 +122,14 @@ func (c *protocolDaoGetStatusContext) PrepareData(data *api.ProtocolDaoStatusRes
 	var err error
 
 	data.IsVotingInitialized = c.node.IsVotingInitialized.Get()
-	if data.IsVotingInitialized {
+
+	if !data.IsVotingInitialized {
+		data.TotalDelegatedVp = nil
```
jshufro, line 127: if `data` is 0-valued when this function is called (and i believe it is) this is a nop and shouldn't be done

jshufro (changes requested):

thomaspanf:
#[276d4ef](https://github.com/rocket-pool/smartnode/pull/599/commits/276d4ef2d715cbd233ac7a1199d981304d1bcba5)

Snapshot isn't on holesky and I don't have a mainnet node so one of you guys should let me know if this works properly. That'd be much appreciated. 

It should look something like this: 
```
Rocket Pool has 1 Snapshot governance proposal(s) being voted on. You have voted on 0 of those. See details using 'rocketpool network dao-proposals'.
```

thomaspanf:
#[6fce08c](https://github.com/rocket-pool/smartnode/pull/599/commits/6fce08cdde6df06fc85e5d741a318efe3b53bacd)
The difference between `types.ResponseStatus_ResourceNotFound` and `types.ResponseStatus_Error` looks like: 

```
tpan@lumiere:~/Downloads$ dev p s

Your Smart Node is currently using the Holesky Test Network.

error during PDAO GetStatus request: route 'pdao/get-status' not found
tpan@lumiere:~/Downloads$ dev p s

Your Smart Node is currently using the Holesky Test Network.

error during PDAO GetStatus request: Network [holesky] does not have a signer registry contract.
```

thomaspanf:
[fee1e68](https://github.com/rocket-pool/smartnode/pull/599/commits/fee1e68ab5909ffd2cdf88ab15a9853718f28183) 

Should `SanitizeHex()` return an error if it isn't 0x prefixed? We should consider Joe's [implementation](https://github.com/thomaspanf/node-manager-core/pull/1), which doesn't require a 0x prefix and only removes it if present. 

We can close [NMC#17](https://github.com/rocket-pool/node-manager-core/pull/17) if this looks good, since all the EIP712 related logic resides in the smartnode now.

shared/utils/input.go
```diff
@@ -21,6 +23,12 @@ const (
 	MinPasswordLength    int = 12
 )
 
+type EIP712Components struct {
+	V uint8    `json:"v"`
+	R [32]byte `json:"r"`
+	S [32]byte `json:"s"`
```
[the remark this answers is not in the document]
thomaspanf: `setSigner()` is ordered v/r/s in `RocketSignerRegistry.sol`

https://github.com/rocket-pool/rocketpool-delegate/blob/4cfd0f5f8616beef312572ba73329d80ecb5152b/src/RocketSignerRegistry.sol#L19

shared/eip712/eip712.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, fmt.Errorf("error sanitizing EIP-712 signature string as hex: %w", err)
+	}
+	if len(decoded) != EIP712Length {
+		return nil, fmt.Errorf("error sanitizing EIP-712 signature string: invalid length %d (expected %d)", len(decoded), EIP712Length)
+	}
+	return decoded, nil
+}
+
+// ParseEIP712Components expects a sanitized 65 byte EIP-712 signature
+// and returns v/r/s: v as uint8 and r, s as [32]byte
```
jshufro, line 50: r/s/v

shared/eip712/eip712.go
```diff
@@ -0,0 +1,69 @@
...
+func ParseEIP712Components(data []byte) (EIP712Components, error) {
+	if len(data) != EIP712Length {
+		return EIP712Components{}, fmt.Errorf("error parsing EIP-712 signature string: invalid length %d (expected %d)", len(data), EIP712Length)
+	}
+
+	var v uint8
+	var r [32]byte
+	var s [32]byte
```
jshufro, line 58: So 0-values for arrays are actually valid `copy` dsts: https://go.dev/play/p/hm7St-bmkqH

that means we can do this instead:
```golang
out := &EIP712Components{
    v: data[64],
}

copy(out.r[:], data[0:32])
copy(out.s[:], data[32:64])

return out, nil
```
which is slightly cleaner than making 3 stack vars

shared/eip712/eip712.go
```diff
@@ -0,0 +1,69 @@
...
+func (e EIP712Components) String() {
+	fmt.Printf("EIP712 Components:\n")
+	fmt.Printf("V: %d\n", e.V)
+	fmt.Printf("R: %x\n", e.R)
+	fmt.Printf("S: %x\n", e.S)
+}
+
+// SanitizeHexInput checks for and strips a 0x prefix, returning an error if one is not found.
+// It returns the byte slice representing the decoded hex, or a decoding error
+func SanitizeHex(inp string) ([]byte, error) {
```
jshufro, line 28: I think we should just use <https://pkg.go.dev/github.com/ethereum/go-ethereum/common/hexutil>

Instead of SanitizeHex we can just call hexutil.Decode and bubble up any errors that occur.

The 0x prefix is mandatory in go-ethereum and for consistency we may as well require it ourselves.

So basically, delete this function

shared/eip712/eip712.go
```diff
@@ -0,0 +1,69 @@
...
+	if !strings.HasPrefix(inp, "0x") {
+		return nil, errors.New("expected hex prefix")
+	}
+	return hex.DecodeString(strings.TrimPrefix(inp, "0x"))
+}
+
+// SanitizeEIP712String checks that a 0x-prefixed hex string is the proper length to be
+// parsed as a EIP-712 signature, and parses it, returning an error if the length is invalid
+// or the hex cannot be decoded.
+func SanitizeEIP712String(inp string) ([]byte, error) {
```
jshufro, line 38: Can we instead use a function with the signature:
`func (out *EIP712String) Decode(inp string) error`
This would replace SanitizeEIP712String and ParseEIP712Components

If we just want to check that a string is valid we can discard the local variable afterwards.

If we actually want to parse and use something, it lets us write this code:
```golang
eip712Components := new(EIP712Components)
err := eip712Components.Decode(inp)
if err != nil {
    return err
}

err = eip712Components.ValidateSignature(msg string)
if err != nil {
    return err
}
```

Which is succinct and go-like.
You will want to implement `Validate()` in this package, as well as `Encode() string` so that you can round-trip a string for testing/printing/logging

rocketpool-daemon/api/pdao/set-signalling-address.go
```diff
@@ -108,12 +106,17 @@ func (c *protocolDaoSetSignallingAddressContext) PrepareData(data *types.TxInfoD
...
+		return types.ResponseStatus_Error, fmt.Errorf("Error sanitizing signature input")
 	}
 
-	data.TxInfo, err = c.registry.SetSigner(c.signallingAddress, opts, sig.V, sig.R, sig.S)
+	components, err := eip712.ParseEIP712Components(signatureSanitized)
+	if err != nil {
+		return types.ResponseStatus_Error, fmt.Errorf("Error parsing EIP-712 components")
+	}
+
+	data.TxInfo, err = c.registry.SetSigner(c.signallingAddress, opts, components.V, components.R, components.S)
```
jshufro, line 119: we shouldn't rely on a contract revert to notify the user of an invalid signature- validate the actual signed message first. you'll have to reconstruct it.

See <https://github.com/Rocket-Rescue-Node/rescue-api/blob/main/util/ethereum.go#L22> for the broad strokes on how to recover the signer (validating the message).

jshufro (changes requested):
the eip712 package is too fundamentally testable to not have a unit test. Please add one :)

shared/eip712/eip712.go
```diff
@@ -0,0 +1,54 @@
...
+	R [32]byte `json:"r"`
+	S [32]byte `json:"s"`
+	V uint8    `json:"v"`
+}
+
+const EIP712Length = 65
+
+// Pretty print for EIP712Components
+func (e EIP712Components) String() {
+	fmt.Printf("EIP712 Components:\n")
```
jshufro, line 19:
```suggestion
	fmt.Println("EIP712 Components:")
```

shared/eip712/eip712.go
```diff
@@ -0,0 +1,54 @@
...
+func (e EIP712Components) String() {
+	fmt.Println("EIP712 Components:")
+	fmt.Printf("R: %x\n", e.R)
+	fmt.Printf("S: %x\n", e.S)
+	fmt.Printf("V: %d\n", e.V)
+}
+
+// Decode decodes a hex-encoded EIP-712 signature string, verifies the length
+// and assigns the appropriate bytes to R/S/V
+func (e *EIP712Components) Decode(inp string) error {
```
jshufro, line 27: mea culpa, let's rename Decode to UnmarshalText and accept inp as []byte
and rename Encode to MarshalText and return ([]byte, error)

so we can fulfill this interface: <https://pkg.go.dev/encoding#TextMarshaler>

shared/eip712/eip712.go
```diff
@@ -0,0 +1,54 @@
...
+type EIP712Components struct {
+	R [32]byte `json:"r"`
+	S [32]byte `json:"s"`
+	V uint8    `json:"v"`
+}
+
+const EIP712Length = 65
+
+// Pretty print for EIP712Components
+func (e EIP712Components) String() {
```
jshufro, line 18: rename this to Print(). Functions named String() should return a string to fulfill <https://pkg.go.dev/fmt#Stringer>

however, i think we want the Stringer interface to be implemented as:
```golang
func (e EIP712Components) String() string {
    out, err := e.MarshalText()
    if err != nil {
        // MarshalText should never return an error
        panic(err)
    }
    return string(out)
}
```
so that the "native" format is the 65 byte 0x-prefixed hex string.

jshufro, line 18:
```suggestion
func (e *EIP712Components) Print() {
```
(renames to Print and changes the receiver to a pointer type to match your other receivers, which is moderately important)

shared/eip712/eip712_test.go
```diff
@@ -0,0 +1,48 @@
...
+	// Encode eip712Components back to a string
+	encodedSig := eip712Components.Encode()
+
+	if encodedSig != signature {
+		t.Fatalf("Expected %s but got %s", signature, encodedSig)
+	}
+
+}
+
+func TestInvalidSignature(t *testing.T) {
```
jshufro, line 30: (as mentioned offline) add a test-case that is an invalid signature but valid hex, ie, the wrong length.

Rename this testcase to TestInvalidEIP712Hex or something, since it doesn't actually fail against an invalid signature.

jshufro (changes requested):
almost there! the general shape is now correct

shared/eip712/eip712.go
```diff
@@ -0,0 +1,84 @@
...
+
+	copy(e.R[:], inp[0:32])
+	copy(e.S[:], inp[32:64])
+	e.V = inp[64]
+
+	return nil
+}
+
+// MarshallText initializes an empty byte slice, copies fields R/S/V into signatureBytes then returns it
+func (e *EIP712Components) MarshallText() ([]byte, error) {
```
jshufro, line 41: `MarshalText`

function name is actually super important, otherwise we aren't implementing the TextMarshaler interface.

shared/eip712/eip712.go
```diff
@@ -0,0 +1,84 @@
...
+// Pretty print for EIP712Components
+func (e *EIP712Components) Print() {
+	fmt.Println("EIP712 Components:")
+	fmt.Printf("R: %x\n", e.R)
+	fmt.Printf("S: %x\n", e.S)
+	fmt.Printf("V: %d\n", e.V)
+}
+
+// UnmarshallText verifies the length of a decoded EIP-712 signature and assigns the appropriate bytes to R/S/V
+func (e *EIP712Components) UnmarshallText(inp []byte) error {
```
jshufro, line 28: `UnmarshalText`

jshufro (approved):

jshufro (changes requested):
woops didn't mean to press approve yet

shared/eip712/eip712.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		// MarshalText should never return an error
+		panic(err)
+	}
+	return fmt.Sprintf("%x", out)
+
+}
+
+// UnmarshalText verifies the length of a decoded EIP-712 signature and assigns the appropriate bytes to R/S/V
+func (e *EIP712Components) UnmarshalText(inp []byte) error {
```
jshufro, line 39: ok this is gonna be moderately confusing, but we should expect `inp` to be a 0x prefixed array of bytes with a length of exactly 130....

UnmarshalText expects "UTF-8-encoded text". That means a string of the format `0x[0-9a-fA-F]{130}`

The `encoding` package also has an `UnmarshalBinary` and `MarshalBinary` interface which is what you actually implemented here.

jshufro (changes requested):

shared/eip712/eip712.go
jshufro: (and MarshalText should do the reverse, of course)

jshufro (approved):

PR #600, opened 2024-07-17
Mark bloXroute Max Profit relay as censoring.

Darkmessage:
Since December 18th 2023, bloXroute Max Profit relay is also censoring OFAC transactions.

This was announced [on Twitter here](https://x.com/bloxroute/status/1736819783520092357) and is now documented [on their website](https://docs.bloxroute.com/apis/mev-solution/mev-relay-for-validators#relay-types):

> Both two relays currently propagate all available transactions/bundles except the ones that interacted with addresses sanctioned by OFAC, and they have the same filtering requirement, although for legacy reasons, bloXroute still maintains the two relay names in parallel.

### DIFF
--- a/shared/services/config/mev-boost-config.go
+++ b/shared/services/config/mev-boost-config.go
@@ -381,8 +381,8 @@ func createDefaultRelays() []config.MevRelay {
 		// bloXroute Max Profit
 		{
 			ID:          config.MevRelayID_BloxrouteMaxProfit,
-			Name:        "bloXroute 1 (formerly Max Profit)",
-			Description: "Select this to enable the relay from bloXroute formerly known as \"max profit\". (Both bloXroute relays propagate the same transactions...)",
+			Name:        "bloXroute Max Profit",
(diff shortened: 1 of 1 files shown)


Darkmessage:
I just adjusted the setting that I found during a quick search.

But since I don't know the SmartNode code, maybe some kind of upgrade step is needed, so that the people who use non-regulated profile mode are up-to-date?
But I guess the current relays are always rebuilt and given to the client when restarting the client containers anyway, right?

Darkmessage:
Should also be updated in the docs on the website:

image

So I also created [a PR](https://github.com/rocket-pool/docs.rocketpool.net/pull/91) for that.

0xfornax (changes requested):
Hey Darkmessage, thanks for the contribution!
I'd just keep current the names instead of using 1 and 2, since their docs still refer to the original names. We can keep the warning that both relays propagate the same txs.

0xfornax (approved):

PR #608, opened 2024-07-22
Pass the selected minipool list, instead of every mp

0xfornax:
Fixes a bug where the close minipool command would iterate over every minipool on a node.


0xfornax:
> Can't we remove `addresses` from the function now?
I think that would fit better on the refactor PR we discussed.

PR #609, opened 2024-07-23
Add dynamic settings / resource loading

jclapis:
This allows the Smart Node to load its network settings and resources from disk, rather than hardcoding them directly into the app. Patches and I have been talking about this for a while and better late than never!

Benefits are:
- Ability to add new network configurations for new networks like Kurtosis
- Ability to modify parameters (like URLs or addresses) for existing networks (e.g, add new `previousRocketRewardsPoolAddresses` directly in the settings file instead of having to build a new SN release with an updated address list).

I did my best to port the structure over from HD, but I'd test this all-the-same just to make sure I didn't miss anything. I haven't actually run this build of the SN in a vacuum to make sure all of the template parameters have been updated so pay particular attention to that. Otherwise, enjoy!


0xfornax:
Closing since we didn't continue with the V2 branch

PR #613, opened 2024-07-23
Update Nethermind config to support passing the PruningMaxDegreeOfParallelism parameter

SolezOfScience:
Addressing https://github.com/rocket-pool/smartnode/issues/597 -- adding support for Nethermind's PruningMaxDegreeOfParallelism parameter.

Also see, https://github.com/rocket-pool/smartnode-install/pull/134 for where this config parameter is used.

### DIFF
--- a/shared/services/config/nethermind-params.go
+++ b/shared/services/config/nethermind-params.go
@@ -134,8 +134,8 @@ func NewNethermindConfig(cfg *RocketPoolConfig) *NethermindConfig {
 		FullPruningMaxDegreeOfParallelism: config.Parameter{
 			ID:                 "fullPruningMaxDegreeOfParallelism",
 			Name:               "Full pruning parallelism",
-			Description:        "The number of threads allocated to concurrently by Nethermind to prune data.",
-			Type:               config.ParameterType_Uint,
+			Description:        "This option will be used to determine the number of threads allocated to concurrently by Nethermind to prune data.",
+			Type:               config.ParameterType_Int,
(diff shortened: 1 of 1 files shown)


shared/services/config/nethermind-params.go
```diff
@@ -128,6 +131,17 @@ func NewNethermindConfig(cfg *RocketPoolConfig) *NethermindConfig {
 			OverwriteOnUpgrade: false,
 		},
 
+		FullPruningMaxDegreeOfParallelism: config.Parameter{
+			ID:                 "fullPruningMaxDegreeOfParallelism",
+			Name:               "Full pruning parallelism",
+			Description:        "The number of threads allocated to concurrently by Nethermind to prune data.",
+			Type:               config.ParameterType_Uint,
```
0xfornax, line 138: -1 is a valid option 
Sem título

shared/services/config/nethermind-params.go
```diff
@@ -128,6 +131,17 @@ func NewNethermindConfig(cfg *RocketPoolConfig) *NethermindConfig {
 			OverwriteOnUpgrade: false,
 		},
 
+		FullPruningMaxDegreeOfParallelism: config.Parameter{
+			ID:                 "fullPruningMaxDegreeOfParallelism",
+			Name:               "Full pruning parallelism",
+			Description:        "The number of threads allocated to concurrently by Nethermind to prune data.",
```
0xfornax, line 137: The description is a bit misleading as the option will determine the number of threads but is not directly that number. e.g. the 0 option doesn't mean 0 threads.

Issue #601, opened 2024-07-18
start-mev-boost.sh doesn't use additional flags


nixorokish:
https://github.com/rocket-pool/smartnode-install/blob/master/install/scripts/start-mev-boost.sh

It is currently impossible to set a `min-bid` in the smart node stack using the "Additional Flags" field

PR #617, opened 2024-08-03
Addressing rocket-pool/smartnode#601 on v2 branch

SolezOfScience:
Porting the changes made as part of https://github.com/rocket-pool/smartnode-install/pull/135 to the v2 branch.

### DIFF
--- a/install/deploy/templates/mev-boost.tmpl
+++ b/install/deploy/templates/mev-boost.tmpl
@@ -19,6 +19,7 @@ services:
       - NETWORK={{.Network}}
       - MEV_BOOST_PORT={{.MevBoost.Port}}
       - MEV_BOOST_RELAYS={{.MevBoost.GetRelayString}}
+      - MEV_BOOST_ADDITIONAL_FLAGS={{.MevBoost.AdditionalFlags}}
     entrypoint: sh
     command: "/usr/share/rocketpool/scripts/{{.GetMevBoostStartScript}}"
     cap_drop:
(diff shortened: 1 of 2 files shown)


jshufro:
I think you accidentally included an extra commit in this PR. It should just be the mev-boost issue, but has a commit intended to implement partial wallet recovery.

SolezOfScience:
You're right - not sure how that happened. I'll fix this.

install/deploy/templates/mev-boost.tmpl
```diff
@@ -19,6 +19,7 @@ services:
       - NETWORK={{.Network}}
       - MEV_BOOST_PORT={{.MevBoost.Port}}
       - MEV_BOOST_RELAYS={{.MevBoost.GetRelayString}}
+      - MEV_BOOST_ADDITIONAL_FLAGS={{.MevBoost.AdditionalFlags}}
```
jshufro, line 22: I don't see this being added to the config struct in this pr. I'm assuming it was already cherry-picked in to v2?

jshufro (approved):

install/deploy/templates/mev-boost.tmpl
SolezOfScience: I think it's already being passed, unless I'm misunderstanding. [Code Link](https://github.com/rocket-pool/smartnode/blob/v2/shared/config/mev-boost-config.go#L73)

Also looks like it gets passed [here](https://github.com/rocket-pool/smartnode/blob/v2/shared/config/mev-boost-config.go#L282). I also checked the UI on my node and confirmed the parameter box is present but it won't be used without this change.

PR #621, opened 2024-08-11
Addressing smartnode-issue-572

SolezOfScience:
### Overview
This is change 1/2 to address smartnode-issue-572. 

This change includes a `key-recovery-manager` class which is a refactored version of the existing `recover-keys`. For the time being, the old version is being kept and will be removed in a follow-up PR which migrates the remaining daemon API implementations to use the new version and support the new CLI parameters.

### Failure Reason Collection
The approach collects errors as it iteratively tries to recover keys. If keys have not been attempted to be processed, you won't get a publicKey -> error association. However, you will get a general error response. Once publicKeys are able to be procured, the response objects are used. 

### Limiting Unexpected State
Overall, this change maintains the status quo for how the API tries to manage/limit risk of unexpected state and how it's documented. At their core, these APIs will always have risk of unexpected state given the serial nature of how keys are written by the keystore manager. Without a true atomic batch-write procedure, there will always be risk of the process leaving unfinished work. This change does not intend to address this, but is acknowledged as a potential improvement in the future should it become worthwhile. 

### Testing
No testing has been run on this change yet. I would prefer to setup unit testing rather than try enacting manual testing. I'll be following up on this topic as the PR receives feedback.


client/wallet.go
```diff
@@ -82,8 +82,8 @@ func (r *WalletRequester) Masquerade(address common.Address) (*types.ApiResponse
 }
 
 // Rebuild the validator keys associated with the wallet
-func (r *WalletRequester) Rebuild() (*types.ApiResponse[api.WalletRebuildData], error) {
-	return client.SendGetRequest[api.WalletRebuildData](r, "rebuild", "Rebuild", nil)
+func (r *WalletRequester) Rebuild(enablePartialRebuild bool) (*types.ApiResponse[api.WalletRebuildData], error) {
+	return client.SendGetRequest[api.WalletRebuildData](r, "rebuild", "Rebuild", map[string]string{"enable-partial-rebuild": strconv.FormatBool(enablePartialRebuild)})
```
jshufro, line 86: interesting. bool types are native json types- do we need to use a string?

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -2,11 +2,11 @@ package wallet
 
 import (
 	"fmt"
+	"github.com/rocket-pool/smartnode/v2/rocketpool-cli/utils"
```
jshufro, line 5: let's keep the imports organized and split according to stdlib vs not

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,52 @@ func rebuildWallet(c *cli.Context) error {
...
 	}
 
+	var enablePartialRebuildValue = false
+	if enablePartialRebuild.Name != "" {
+		enablePartialRebuildValue = c.Bool(enablePartialRebuild.Name)
+	}
+
 	// Log
 	fmt.Println("Rebuilding node validator keystores...")
+	fmt.Printf("Partial rebuild enabled: %s.\n", enablePartialRebuild.Name)
```
jshufro, line 64: `enablePartialRebuild.Name` here is equal to `enable-partial-rebuild` per https://github.com/rocket-pool/smartnode/blob/9bd551093d9de7eddb5cc7f985989bdf18c8b585/rocketpool-cli/commands/wallet/utils.go#L61

Did you mean to printf the value of `enablePartialRebuildValue` instead?

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,52 @@ func rebuildWallet(c *cli.Context) error {
...
 
 	// Rebuild wallet
-	response, err := rp.Api.Wallet.Rebuild()
-	if err != nil {
-		return err
+	response, _ := rp.Api.Wallet.Rebuild(enablePartialRebuildValue)
```
jshufro, line 67: we still need to catch and return the error here, as there can be http failures between the cli and the api

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,52 @@ func rebuildWallet(c *cli.Context) error {
...
+			fmt.Printf("Public Key: %s - Failure Reason: %s\n", pubkey.Hex(), reason)
+		}
+	} else {
+		fmt.Println("No failures reported.")
 	}
 
-	// Log & return
-	fmt.Println("The node wallet was successfully rebuilt.")
-	if len(response.Data.ValidatorKeys) > 0 {
+	fmt.Println("The response for rebuilding the node wallet was successfully received.")
```
jshufro, line 79: this is a bit too verbose to expose to the user- they don't need to know about the request/response nature of their interactions.

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,52 @@ func rebuildWallet(c *cli.Context) error {
...
+	if len(response.Data.RebuiltValidatorKeys) > 0 {
 		fmt.Println("Validator keys:")
-		for _, key := range response.Data.ValidatorKeys {
+		for _, key := range response.Data.RebuiltValidatorKeys {
 			fmt.Println(key.Hex())
```
jshufro, line 83:
```suggestion
		for i, key := range response.Data.RebuiltValidatorKeys {
			fmt.Printf("\t%d.%s", i+1, key.Hex())
```

bit more ergonomic to enumerate the returned keys (starting with 1) so users can quickly see that the expected number of validators were recovered.

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,52 @@ func rebuildWallet(c *cli.Context) error {
...
 
 	// Rebuild wallet
-	response, err := rp.Api.Wallet.Rebuild()
-	if err != nil {
-		return err
+	response, _ := rp.Api.Wallet.Rebuild(enablePartialRebuildValue)
+
+	// Handle and print failure reasons with associated public keys
+	if len(response.Data.FailureReasons) > 0 {
+		fmt.Println("Failure reasons:")
```
jshufro, line 71: before this, print a blurb like:
"Some keys could not be recovered. You may need to import them manually, as they are not associated with your node wallet mnemonic. See the documentation for more details."

rocketpool-cli/commands/wallet/utils.go
```diff
@@ -57,6 +57,11 @@ var (
 		Aliases: []string{"a"},
 		Usage:   "If you are recovering a wallet that was not generated by the Smartnode and don't know the derivation path or index of it, enter the address here. The Smartnode will search through its library of paths and indices to try to find it.",
 	}
+	enablePartialRebuild = &cli.StringSliceFlag{
+		Name:    "enable-partial-rebuild",
+		Aliases: []string{"epr"},
```
jshufro, line 62: for flags we should avoid trigraph shorthand. maybe just "p" here for partial?

rocketpool-daemon/common/validator/key-recovery-manager/dry-run-key-recovery-manager.go
```diff
@@ -0,0 +1,149 @@
...
+	recoveredKeys := make(map[beacon.ValidatorPubkey]bool)
+	recoveryFailures := make(map[beacon.ValidatorPubkey]error)
+	var passwords map[string]string
+
+	keyFiles, err := d.manager.LoadFiles()
+	if err != nil {
+		return recoveredKeys, recoveryFailures, err
+	}
+
+	if len(keyFiles) > 0 {
```
jshufro, line 63: check `if len(keyFiles) == 0` and return early. Outdent the entire conditional below.

rocketpool-daemon/common/validator/key-recovery-manager/key-recovery-manager.go
```diff
@@ -0,0 +1,14 @@
+package key_recovery_manager
+
+import (
+	"github.com/rocket-pool/node-manager-core/beacon"
+)
+
+type KeyRecoveryManager interface {
+	RecoverMinipoolKeys() ([]beacon.ValidatorPubkey, map[beacon.ValidatorPubkey]error, error)
```
jshufro, line 8: can this just be argumented with `enablePartialRecovery bool` instead of having two different implementations? Might reduce repeated code.

rocketpool-daemon/common/validator/validator-manager.go
```diff
@@ -3,21 +3,33 @@ package validator
...
 	"github.com/rocket-pool/rocketpool-go/v2/rocketpool"
 	"github.com/rocket-pool/smartnode/v2/rocketpool-daemon/common/utils"
 	"github.com/rocket-pool/smartnode/v2/shared/config"
+	"github.com/rocket-pool/smartnode/v2/shared/types/api"
 	eth2types "github.com/wealdtech/go-eth2-types/v2"
 	types "github.com/wealdtech/go-eth2-types/v2"
+	eth2ks "github.com/wealdtech/go-eth2-wallet-encryptor-keystorev4"
+	"gopkg.in/yaml.v3"
+	"os"
+	"path/filepath"
```
jshufro, line 24: let's split the imports up into std vs not

rocketpool-daemon/common/validator/validator-manager.go
```diff
@@ -272,3 +284,139 @@ func (m *ValidatorManager) checkIfReady() error {
...
+	if err != nil {
+		return status, err
+	}
+	if !walletcore.IsWalletReady(status) {
+		return status, fmt.Errorf("wallet is not ready")
+	}
+	return status, nil
+}
+
+func (m *ValidatorManager) InitializeBindings(status walletcore.WalletStatus) (*node.Node, *minipool.MinipoolManager, error) {
```
jshufro, line 299: this function returns a `node.Node` and a `minipool.MinipoolManager`, which are then passed in to `GetMinipools`.

This seems like a poor design choice: why not simply create those pointers in `NewValidatorManager` and store them as unexported fields on `ValidatorManager`? They can then be accessed in `GetMinipools` with `m.node` or `m.minipoolManager`

client/wallet.go
bdchatham: I believe I was just following convention on this one. Looked like there were multiple cases in the same class of using the same pattern. 

`skip-validator-key-recovery` is one example I'm seeing immediately.

SolezOfScience: I believe I was just following convention on this one. For example, the `skip-validator-key-recovery` parameter uses a similar code flow to translate the string to boolean. 

I could see the perspective that this is a bit silly since the parameter is a string, gets converted to a bool to call the method then is immediately converted back to a string for the payload. I also cringe at the idea of passing a string around which is serving as a boolean, so neither way is elegant.

rocketpool-cli/commands/wallet/rebuild.go
SolezOfScience: Thanks for pointing this out. Frankly I found the use of these oddly confusing. For example, recover.go, [it checks if the name is non-empty before using it](https://github.com/rocket-pool/smartnode/blob/v2/rocketpool-cli/commands/wallet/recover.go#L59) but given the definition of the flags, I don't see how it could be empty.

Regardless, I'll definitely update this log.

rocketpool-daemon/common/validator/key-recovery-manager/key-recovery-manager.go
SolezOfScience: Discussed in discord. Would like to avoid parameterizing the condition. Will explore multiple methods approach.

client/wallet.go
SolezOfScience: The SendGetRequest method takes a map[string]string. Refactoring this down could make this more generic, but for now requires parameters as strings.

rocketpool-cli/commands/wallet/rebuild.go
jshufro: Well that looks like a bug :(

rocketpool-cli/commands/wallet/rebuild.go
```diff
@@ -54,41 +54,56 @@ func rebuildWallet(c *cli.Context) error {
...
+		fmt.Println("Restarting Validator Client...")
+		_, err = rp.Api.Service.RestartVc()
+		if err != nil {
+			fmt.Printf("Error restarting Validator Client: %s\n", err.Error())
+			fmt.Println("Please restart the Validator Client manually at your earliest convenience to load the keys.")
+			return nil
+		}
+		fmt.Println("Validator Client restarted successfully.")
+	} else {
+		fmt.Println("No validator keys were found.")
```
jshufro, line 105: I have a slight preference to see this moved up to L84 and the conditional inverted. An early return can prevent the need for an `else` block.

jshufro (approved):
approved, but i think we need to do some thorough testing before we land

rocketpool-cli/commands/wallet/rebuild.go
SolezOfScience: Gotcha - once you left that comment I figured that may be the case. I'll take a note.

0xfornax:
> No testing has been run on this change yet. I would prefer to setup unit testing rather than try enacting manual testing. I'll be following up on this topic as the PR receives feedback.

Hey @SolezOfScience, thanks for the PR! Are you still considering unit tests?

SolezOfScience:
Hey @0xfornax, sorry for the delay. A change in my professional role has left me with less bandwidth to finish this than I originally expected. The next few weeks are very busy for me between going to Pragma and ETHGlobal and some personal life commitments. It is probably best to have the rest of the testing picked up by someone from the team.

Armanidashh (approved):

0xfornax:
Closing as it targets v2.

PR #626, opened 2024-08-14
V1 auto init vp

thomaspanf:


jshufro:
I think the manual testing you already did will have to suffice. I want to have a testing framework for each of the 'background' tasks that rocketpool_node handles in v2, but i don't want to write it for v1.

jshufro (approved):

PR #643, opened 2024-09-02
Remove the Eden relay.

0xfornax:
Considering this [announcement](https://x.com/titanrelayxyz/status/1829202005354807374) the Eden relay will be removed from the next release


thomaspanf (commented):
Don't we need to remove these too? 

https://github.com/rocket-pool/smartnode/blob/remove-eden-relay/shared/types/config/types.go#L98

https://github.com/rocket-pool/smartnode/blob/remove-eden-relay/shared/services/config/mev-boost-config.go#L395

https://github.com/rocket-pool/smartnode/blob/remove-eden-relay/rocketpool-cli/service/config/settings-mev-boost.go#L173

Issue #634, opened 2024-08-29
[V2] Segfault on setting RPL withdrawal address


thomaspanf:
`rp n srwa`

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5c4458]

goroutine 1 [running]:
math/big.(*Int).Cmp(0xab0ab7226b769ac?, 0xb4284f4af73718b5?)
    math/big/int.go:381 +0x18
github.com/rocket-pool/smartnode/v2/rocketpool-cli/commands/node.setRplWithdrawalAddress(0xc000596180, {0x7ffccac8e188, 0x2a})
    github.com/rocket-pool/smartnode/v2/rocketpool-cli/commands/node/set-rpl-withdrawal-address.go:122 +0x969

PR #644, opened 2024-09-03
Fixed segfault by populating nil pointer

thomaspanf:
Addresses #634 

`rp-cli` segfaults [here](https://github.com/rocket-pool/smartnode/blob/76440160cd8fbbb49927db0b8247dd5e6defce85/rocketpool-cli/commands/node/set-rpl-withdrawal-address.go#L122) because`rp-daemon` never populates `response.Data.RplStake`


jshufro (approved):
Increasingly disenchanted with the framework. This landmine keeps coming up.

PR #655, opened 2024-09-23
Handle a case where priorityFee > maxFee in the task loop

thomaspanf:
Automatic transactions in the task loop fail when the user-inputted `priorityFee` is greater than the oracle based `maxFee`. This PR addresses this case by setting `priorityFee` to `min(priorityFee, 25% of the oracle based maxFee`) when `priorityFee > maxFee` for transactions in the node task loop. 

For example: automatically initializing vote power, defending pDAO proposals, bond reductions and so on. 

The change is also applied to watchtower task `process-penalties.go`

### DIFF
--- a/rocketpool/node/node.go
+++ b/rocketpool/node/node.go
@@ -394,6 +394,24 @@ func updateNetworkState(m *state.NetworkStateManager, log *log.ColorLogger, node
 	return state, totalEffectiveStake, nil
 }
 
+// Checks if the user-inputted priorityFee is greater than the oracle based maxFee
+// If so, return the min(priorityFee, 25% of the oracle based maxFee)
+func GetPriorityFee(priorityFee *big.Int, maxFee *big.Int) *big.Int {
+	// Check if priorityFee is less than maxFee
+	if priorityFee.Cmp(maxFee) < 0 {
+		return priorityFee
+	}
+
+	quarterMaxFee := new(big.Int).Div(maxFee, big.NewInt(4))
+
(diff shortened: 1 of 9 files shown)


rocketpool/node/node.go
```diff
@@ -394,6 +394,24 @@ func updateNetworkState(m *state.NetworkStateManager, log *log.ColorLogger, node
...
+	}
+
+	quarterMaxFee := new(big.Int).Div(maxFee, big.NewInt(4))
+
+	// Gets the min(priorityFee, 25% of the oracle based maxFee)
+	if priorityFee.Cmp(quarterMaxFee) < 0 {
+		return priorityFee
+	} else {
+		return quarterMaxFee
+	}
```
jshufro, line 412:
```suggestion
	return quarterMaxFee
```

jshufro (changes requested):
This is _fine_ but we can do better.

Every `t *taskThing` has a maxPriorityFee setting. Instead, let's make every `t *taskThing` have a `gasSettings *GasSettings` field. Put `maxPriorityFee` on `GasSettings`, get rid of GetPriorityFee and instead have a function called something like `func (g *GasSettings) ApplyMaxFeeTo(opts *bind.CallOpts)` which sets opts.GasFeeCap, opts.GasTipCap, and opts.GasLimit.

Then we get a lot of deduplication out of this PR. Also, add a unit test? it's all arithmetic and logic that is very easy to unit test.

jshufro (approved):
Edit: just realized this is on v1. Fine to merge as-is I guess, but make sure we do this right when we apply it to v2

PR #658, opened 2024-09-25
Delete old treegens

jshufro:
There's too many to update on refactor.

Delete the unused ones, and plan to reinstate them when the architecture can be DRYer.


shared/services/rewards/types.go
```diff
@@ -14,6 +14,8 @@ import (
 type rewardsFileVersion uint64
 
 const (
+	FarEpoch uint64 = 18446744073709551615
```
jshufro, line 17: generator-impl-v2 previously defined this, but it is used in several implementations.

jshufro: ... and it has to be in a separate `const` block from the enum, or it takes a position

PR #672, opened 2024-10-11
Enable MEV-Boost for the Holesky network.

0xfornax:


jclapis:
Everyone in the Constellation cohort thanks you for this :100:

PR #708, opened 2024-11-27
V1 Update Metrics Exporter and deprecate SnapshotDelegation

thomaspanf:
- Removed all traces of `SnapshotDelegation`
- Updated the `metrics-exporter.go` and `snapshot-collector.go` to support the new voting system. 

 These are the values passed into Grafana: 
```
	channel <- collector.activeProposals
	channel <- collector.closedProposals
	channel <- collector.votesActiveProposals
	channel <- collector.votesClosedProposals
	channel <- collector.nodeVotingPower
	channel <- collector.delegateVotingPower
```

No changes were needed for `activeProposals` and `closedProposals`.  `votesActiveProposals` and `votesClosedProposals` should display the correct info now that `collector.delegateAddress` is replaced with `collector.signallingAddress`. 

We use `propMgr.GetArtifactsForVoting(blockNumber, address)` and pass the node address or on-chain delegate address to grab `nodeVotingPower` and `delegateVotingPower`, similarly to how it's done in `rp pdao status`

One thing we could do is remove the `if time.Since(timestamp)>= hoursToWait` check on the voting power calls and have it run every 5 minutes alongside the task loop. It doesn't make calls to the Snapshot API so I think this is okay.


0xfornax:
> One thing we could do is remove the `if time.Since(timestamp)>= hoursToWait` check on the voting power calls and have it run every 5 minutes alongside the task loop. It doesn't make calls to the Snapshot API so I think this is okay.

I think it could be confusing to have a few fields of the Governance tab update faster and others take hours. Think it's ok to keep it as it is.

0xfornax (approved):

PR #712, opened 2024-12-05
Add suggested block gas limit parameter

0xfornax:


jshufro:
Couple things:

It's a _gas_ limit, not a block limit (see lh docs here)
```
      --gas-limit <INTEGER>
          The gas limit to be used in all builder proposals for all validators
          managed by this validator client. Note this will not necessarily be
          used if the gas limit set here moves too far from the previous block's
          gas limit. [default: 30,000,000]
```

And, the beacon-apis spec says this about MEV-boost:
```
Prepares the beacon node for engaging with external builders. The information must be sent by the beacon node to the builder network. It is expected that the validator client will send this information periodically to ensure the beacon node has correct and timely registration information to provide to builders. The validator client should not sign blinded beacon blocks that do not adhere to their latest fee recipient and gas limit preferences.

Note that only registrations for active or pending validators must be sent to the builder network. Registrations for unknown or exited validators must be filtered out and not sent to the builder network.```

Which implies that the value is not just used for locally built blocks, but part of the registration that is sent to relays, and while it may not be respected, it is still available to builders.

0xfornax:
> Which implies that the value is not just used for locally built blocks, but part of the registration that is sent to relays, and while it may not be respected, it is still available to builders.

I understand the EC value would be used if producing a block locally and the BN/VC value when externally building it.

PR #764, opened 2025-03-21
Issue 731 gas limit update

### LINKED ISSUE #731: Gas Limit Update
Briefly:

- The community wanted to customize their gas limit votes so they could support some value other than the default 30m that the clients use.
- The community asked us to make the gas limit customizable, and we added a field to the user-settings yml that accepts a gas limit to use, with a default value of 30m to match the clients
- Some members of the community increased this value to various higher values, with 60m and 36m both being somewhat popular.
- Some members of the community may have deliberately left or set the value to 30m
- See @nixorokish tweet here <https://x.com/nixorokish/status/1881961457988018650>
- The community would now like us to increase the default to 36m

This is unfortunately fraught. We can't check if the value is set to 30m and overwrite with 36m because a node operator may have deliberately voted for 30m, and we'd erase their vote (really bad vibes for that NO). Unfortunately the way we handle yml doesn't differentiate between a value set by the user, or a default value generated on install/upgrade.

The best solution in my opinion is to add a check to `rocketpool service install`- if gas is set to 30m, ask the user if they would like to accept a new default of 36m as recommended by the maintainers (i think it's ok for us to have an opinion here?)- or to keep their setting at 30m, and include a link to an explainer on what the gas limit is/means.

I am writing this ticket because it needs to be triaged/prioritized by the team.

0xfornax:
Close  #731


thomaspanf (approved):

jshufro (changes requested):

shared/services/config/migration/v1153-manager.go
```diff
@@ -0,0 +1,26 @@
...
+		return fmt.Errorf("expected a section called `executionCommon` but it didn't exist")
+	}
+
+	config, exists := executionCommonSettings["suggestedBlockGasLimit"]
+	if !exists {
+		return nil
+	}
+
+	// If using the previous or current defaults, change to the new one
+	if config == "30000000" || config == "36000000" {
```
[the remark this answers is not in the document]
0xfornax: The first version had a prompt, but it surfaced another issue as this code would be called multiple times for a single migration.  So a proper fix would involve a bigger refactor and I don't think it makes sense for us to continue using the 30M limit on the next release.

shared/utils/cli/prompt/prompt.go
```diff
@@ -0,0 +1,88 @@
...
+	selectedIndex := index - 1
+	selectedOption := options[selectedIndex]
+
+	// Return
+	return selectedIndex, selectedOption
+
+}
+
+// Prompts the user to verify that there is nobody looking over their shoulder before printing sensitive information.
+func ConfirmSecureSession(warning string) bool {
```
jshufro, line 81: This looks unused based on this PR- if we copied this in from NMC we should simultaneously remove references to the old ConfirmSecureSession

jshufro (approved):

shared/utils/cli/prompt/prompt.go
0xfornax: Not from the NMC. I'm moving from the cli/prompt.go. I'm going ahead and updating references everywhere on this PR.

PR #775, opened 2025-03-27
Use the last submission block as target if the node didn't participate in consensus

0xfornax:
This PR makes the RPL price submission work as a health check for oDAO nodes. If a node didn't participate in the previous round, they're going to use that past block as the submission target.

### DIFF
--- a/rocketpool/watchtower/submit-rpl-price.go
+++ b/rocketpool/watchtower/submit-rpl-price.go
@@ -381,8 +381,12 @@ func (t *submitRplPrice) run(state *state.NetworkState) error {
 
 	// Check if the node has submitted prices for the latest block
 	if lastSubmissionBlock != 0 {
+		// Create context with timeout
+		ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
+		defer cancel()
+
 		// Get the lastSubmissionBlock timestamp
-		lastSubmissionBlockHeader, err = t.rp.Client.HeaderByNumber(context.Background(), big.NewInt(int64(lastSubmissionBlock)))
+		lastSubmissionBlockHeader, err = t.rp.Client.HeaderByNumber(ctx, big.NewInt(int64(lastSubmissionBlock)))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


rocketpool/watchtower/submit-rpl-price.go
```diff
@@ -367,15 +369,46 @@ func (t *submitRplPrice) run(state *state.NetworkState) error {
...
-		return err
+	var hasSubmittedPastBlock bool
+	var nextSubmissionTime time.Time
+	var targetBlockNumber uint64
+	var lastSubmissionBlockHeader *types.Header
+
+	// Check if the node has submitted prices for the latest block
+	if lastSubmissionBlock != 0 {
+		// Get the lastSubmissionBlock timestamp
+		lastSubmissionBlockHeader, err = t.rp.Client.HeaderByNumber(context.Background(), big.NewInt(int64(lastSubmissionBlock)))
```
jshufro, line 385: We _should_ be using contexts with sensible timeouts.

rocketpool/watchtower/submit-rpl-price.go
```diff
@@ -367,15 +369,46 @@ func (t *submitRplPrice) run(state *state.NetworkState) error {
...
+			return err
+		}
+		hasSubmittedPastBlock, err = t.hasSubmittedSpecificBlockPrices(nodeAccount.Address, lastSubmissionBlock, lastSubmissionBlockHeader.Time, state.NetworkDetails.RplPrice)
+		if err != nil {
+			t.log.Printlnf("Error checking if node has submitted prices for block %d: %s", lastSubmissionBlock, err.Error())
+			return err
+		}
+	}
+
+	if !hasSubmittedPastBlock {
```
jshufro, line 397: you can move the `else` part of this conditional to the above conditional, because it only runs if `hasSubmittedSpecificBlockPrices` returns true there. Then you can change this conditional to `if targetBlockNumber == 0`, which is a bit more semantic

jshufro (approved):

PR #776, opened 2025-03-31
Fix to find the apporximator for the first interval

0xfornax:
We're having issues on Hoodi to find the approximator impl for interval 0.


shared/services/rewards/generator.go
```diff
@@ -198,7 +198,7 @@ func NewTreeGenerator(logger *log.ColorLogger, logPrefix string, rp RewardsExecu
 			t.generatorImpl = info.generator
 			foundGenerator = true
 		}
-		if !foundApproximator && startInterval < t.index {
+		if !foundApproximator && (startInterval < t.index || t.index == 0) {
```
jshufro, line 201: Can we just make this `<=` like the generator now? Using the previous ruleset was always a bug. Update the comment on L189-89 too?

PR #779, opened 2025-04-15
Add wallet masquerade and end-masquerade commands

thomaspanf:
This PR introduces two commands:

-  `rocketpool wallet masquerade - Change your node's effective address to a different one. Your node will not be able to submit transactions or sign messages since you don't have the corresponding wallet's private key.`
-  `rocketpool wallet end-masquerade - End a masquerade, restoring your node's effective address back to your wallet address if one is loaded.`

While masquerading, the node and watchtower daemon will continue to submit tasks using the initialized (stored on disk) wallet.

Additionally, the `NewWallet` constructor in the wallet package has been refactored to return an interface rather than a concrete struct, allowing flexibility for future wallet implementations.


shared/types/api/wallet.go
```diff
@@ -23,6 +23,9 @@ type WalletStatusResponse struct {
 	PasswordSet       bool           `json:"passwordSet"`
 	WalletInitialized bool           `json:"walletInitialized"`
 	AccountAddress    common.Address `json:"accountAddress"`
+	NodeAddress       common.Address `json:"nodeAddress"`
+	HasAddress        bool           `json:"hasAddress"`
```
0xfornax, line 27: What's the use for HasAddress?

thomaspanf: Ah this is something I forgot to clean up. Thanks for catching it.

rocketpool-cli/wallet/status.go
```diff
@@ -33,13 +34,27 @@ func getStatus(c *cli.Context) error {
...
-		fmt.Println("The node wallet has not been initialized.")
+	// Masquerading
+	emptyAddress := common.Address{}
+	if status.IsMasquerading && status.NodeAddress != emptyAddress {
+		fmt.Printf("The node wallet is initialized, but you are currently masquerading as %s%s%s\n", colorBlue, status.AccountAddress, colorReset)
+		fmt.Printf("Wallet Address: %s\n", status.NodeAddress)
+		fmt.Printf("%sDue to this mismatch, the node cannot submit transactions. Use the command 'rocketpool wallet end-masquerade' to end masquerading and restore your wallet address.%s", colorYellow, colorReset)
+	} else if status.IsMasquerading {
+		fmt.Printf("The node wallet has not been initialized, but you are currently masquerading as %s%s%s\n", colorBlue, status.AccountAddress, colorReset)
+		fmt.Printf("%sThe node cannot submit transactions. Use the command 'rocketpool wallet end-masquerade' to end masquerading.%s", colorYellow, colorReset)
```
jshufro, line 45:
```suggestion
	if status.IsMasquerading {
		if status.NodeAddress != emptyAddress {
			fmt.Printf("The node wallet is initialized, but you are currently masquerading as %s%s%s\n", colorBlue, status.AccountAddress, colorReset)
			fmt.Printf("Wallet Address: %s\n", status.NodeAddress)
			fmt.Printf("%sDue to this mismatch, the node cannot submit transactions. Use the command 'rocketpool wallet end-masquerade' to end masquerading and restore your wallet address.%s", colorYellow, colorReset)
		} else {
			fmt.Printf("The node wallet has not been initialized, but you are currently masquerading as %s%s%s\n", colorBlue, status.AccountAddress, colorReset)
			fmt.Printf("%sThe node cannot submit transactions. Use the command 'rocketpool wallet end-masquerade' to end masquerading.%s", colorYellow, colorReset)
		}
```

Normally I'm a fan of being as outdented as possible, but not if it requires duplicate conditionals between an `if` and `else if`

rocketpool-cli/wallet/status.go
```diff
@@ -33,13 +34,27 @@ func getStatus(c *cli.Context) error {
...
 	}
+
+	// Not Masquerading
+	if !status.IsMasquerading && status.WalletInitialized {
+		fmt.Println("The node wallet is initialized")
+		fmt.Printf("Wallet Address: %s", status.AccountAddress)
+	}
+	if !status.IsMasquerading && !status.WalletInitialized {
+		fmt.Print("The node wallet has not been initialized.")
+	}
```
jshufro, line 55: If you accept the above suggestion, move all of this into an `else` conditional and remove the `!status.IsMasquerading` bits

shared/types/api/wallet.go
```diff
@@ -23,6 +23,9 @@ type WalletStatusResponse struct {
 	PasswordSet       bool           `json:"passwordSet"`
 	WalletInitialized bool           `json:"walletInitialized"`
 	AccountAddress    common.Address `json:"accountAddress"`
+	NodeAddress       common.Address `json:"nodeAddress"`
```
jshufro, line 26: AccountAddress and NodeAddress should have comments that explain the difference

shared/services/wallet/wallet.go
```diff
@@ -75,12 +114,38 @@ type walletStore struct {
 }
 
 // Create new wallet
-func NewWallet(walletPath string, chainId uint, maxFee *big.Int, maxPriorityFee *big.Int, gasLimit uint64, passwordManager *passwords.PasswordManager) (*Wallet, error) {
+func NewWallet(addressPath string, walletPath string, chainId uint, maxFee *big.Int, maxPriorityFee *big.Int, gasLimit uint64, passwordManager *passwords.PasswordManager, addressManager *AddressManager) (Wallet, error) {
+
+	// Check if the address file exists
+	_, err := os.Stat(addressPath)
+	masquerading := !os.IsNotExist(err)
```
jshufro, line 121: I'd move this under the conditional at L123-5, it's always good to put error handling right after the line that returns an error.

shared/services/wallet/masquerade-wallet.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// Reloads wallet from disk
+func (w *masqueradeWallet) Reload() error {
+	_, err := w.loadStore()
+	return err
+}
+
+// Load the wallet store from disk and decrypt it
+func (w *masqueradeWallet) loadStore() (bool, error) {
```
jshufro, line 195: hmm, does this ever get called? Seems like we should probably just return an error here- can't imagine why we'd want to read the wallet store while masquerading.

shared/services/wallet/masquerade-wallet.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// Recover a wallet from a mnemonic
+func (w *masqueradeWallet) Recover(derivationPath string, walletIndex uint, mnemonic string) error {
+	return ErrIsMasquerading
+
+}
+
+// Recover a wallet from a mnemonic - only used for testing mnemonics
+func (w *masqueradeWallet) TestRecovery(derivationPath string, walletIndex uint, mnemonic string) error {
```
jshufro, line 136: i think ErrIsMasquerading is also appropriate here. We don't let you call Recover while masquerading, so TestRecovery should also fail

shared/services/wallet/masquerade-wallet.go
jshufro: Sounds like a reasonable vector to simplify this a bit, right?

PR #785, opened 2025-04-28
refactor: use the built-in max/min to simplify the code

careworry:
In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.


0xfornax (changes requested):
There are some compilation errors.

careworry:
> There are some compilation errors.

Thanks. Modified, Please review again.

PR #786, opened 2025-04-29
Process committees correctly post-electra

jshufro:


shared/services/beacon/client.go
```diff
@@ -94,6 +97,26 @@ type AttestationInfo struct {
 	AggregationBits bitfield.Bitlist
 	SlotIndex       uint64
 	CommitteeIndex  uint64
+	CommitteeBits   bitfield.Bitlist
+}
+
+// if !attestation.ValidatorAttested(i, position, slotInfo.CommitteeSizes) {
```
jshufro, line 103: delet this please before merging @0xfornax

shared/services/beacon/client.go
```diff
@@ -94,6 +97,26 @@ type AttestationInfo struct {
...
+func (a AttestationInfo) ValidatorAttested(committeeIndex int, position int, committeeSizes map[uint64]int) bool {
+	// Calculate the offset in aggregation_bits
+	var offset int
+	if a.CommitteeBits == nil {
+		offset = position
+	} else {
+		committeeOffset := 0
+		if committeeIndex > 0 {
```
jshufro, line 111: this conditional has no effect, remove it and outdent the body

shared/services/rewards/generator-impl-v8.go
```diff
@@ -852,50 +852,58 @@ func (r *treeGeneratorImpl_v8) checkDutiesForSlot(attestations []beacon.Attestat
...
 			continue
 		}
-		rpCommittee, exists := slotInfo.Committees[attestation.CommitteeIndex]
-		if !exists {
-			continue
+		var committeeIndexes []int
+		if attestation.CommitteeBits == nil {
+			committeeIndexes = []int{int(attestation.CommitteeIndex)}
+		} else {
+			committeeIndexes = attestation.CommitteeBits.BitIndices()
```
jshufro, line 859: We should probably just write a `func (a AttestationInfo) GetCommitteeIndices() []int` and move this in there so that we don't muddy up this code

0xfornax (approved):

PR #796, opened 2025-05-12
Fix sending custom tokens without 18 decimals

### LINKED ISSUE #795: `node send` assumes 18 decimals
The `node send` command does not fetch the number of decimals from the contract when transferring ERC20 tokens and uses the default utility function for ETH to convert from float to wei. This works fine for RPL and many other tokens, but e.g. `1 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`, i.e. an intended transfer of 1 USDC is incorrectly converted to 1,000,000,000,000 USDC.

0xfornax:
Closes #795


rocketpool/api/node/send.go
```diff
@@ -79,6 +79,10 @@ func canNodeSend(c *cli.Context, amountWei *big.Int, token string, to common.Add
 	if strings.HasPrefix(token, "0x") {
 		tokenAddress := common.HexToAddress(token)
 
+		if to == tokenAddress {
```
jshufro, line 82: `strings.EqualFold()` here, so you don't have case sensitivity.

shared/types/api/node.go
```diff
@@ -403,7 +403,7 @@ type CreateVacantMinipoolResponse struct {
 type CanNodeSendResponse struct {
 	Status              string             `json:"status"`
 	Error               string             `json:"error"`
-	Balance             *big.Int           `json:"balance"`
+	Balance             float64            `json:"balance"`
```
jshufro, line 406: We should really switch to something like this at some point:
<https://pkg.go.dev/github.com/shopspring/decimal>

jshufro (approved):

thomaspanf (approved):

rocketpool/api/node/send.go
jshufro: Just noticed these are common.Address, which can be compared with `bytes.Equal()`. Given that it's [40]byte under the hood, `==` technically should also work, but I don't like it because if we change them to pointers it breaks

PR #800, opened 2025-05-13
Remove invocations to nethermind prune starter container, preferring docker-based curl

jshufro:
Use curlimages/curl to begin nethermind prunes instead of a bespoke container.


0xfornax:
I see this instead of the version
```
❯ rp s i -d

The Rocket Pool service will be installed --Version: 
```

0xfornax (approved):

PR #834, opened 2025-05-23
Single module monorepo

jshufro:
At some point I soured on go.work. I think there's a variety of reasons, but mainly, our code needs simply aren't complicated enough to need to maintain separate dependency lists.

Also, there is this:
> It is generally inadvisable to commit go.work files into version control systems, for two reasons:
>
> A checked-in go.work file might override a developer’s own go.work file from a parent directory, causing confusion when their use directives don’t apply.
> A checked-in go.work file may cause a continuous integration (CI) system to select and thus test the wrong versions of a module’s dependencies. CI systems should generally not be allowed to use the go.work file so that they can test the behavior of the module as it would be used when required by other modules, where a go.work file within the module has no effect.


.github/workflows/build.yml
jshufro, line 1: Building natively helps cover cases where cross compilation fails
Building in docker is largely just done to ensure that the docker build itself doesn't break.

0xfornax (approved):

PR #839, opened 2025-05-28
Fix for entrypoint with prysm hybrid mode

jshufro:


shared/services/config/rocket-pool-config.go
```diff
@@ -1383,6 +1383,14 @@ func (cfg *RocketPoolConfig) GetMevBoostOpenPorts() string {
 	return fmt.Sprintf("\"%s\"", portMode.DockerPortMapping(port))
 }
 
+// Used by text/template to select an entrypoint based on which consensus client is used.
+func (cfg *RocketPoolConfig) GetEth2Entrypoint() string {
+	if client, _ := cfg.GetSelectedConsensusClient(); client == config.ConsensusClient_Prysm {
```
jshufro, line 1388: this function checks if we're in hybrid mode

PR #847, opened 2025-06-04
Update vc image regex to be less dangerous

jshufro:
See here for behavioral change: https://go.dev/play/p/1IBbAthKGW-
```
Old:
sigp/lighthouse:v110 -> lighthouse
gcr.io/offchainlabs/prysm/validator:v6.0.3 -> validator

New:
sigp/lighthouse:v110 -> sigp/lighthouse
gcr.io/offchainlabs/prysm/validator:v6.0.3 -> gcr.io/offchainlabs/prysm/validator
```

Basically, when we switched to the upstream prysm containers, a NO noticed that it triggered their 15 minute cooldown. This happened because the regex went from extracting `prysm` from `rocketpool/prysm` to extracting `validator` from `gcr.io/offchainlabs/prysm/validator:v6.0.3`

This is _bad_. Semantically, everything preceding the `:` is relevant to the specific image being used, and if orgs change their tag locations, we should still sit out for 15 minutes to be safe.

### DIFF
--- a/rocketpool-cli/service/service.go
+++ b/rocketpool-cli/service/service.go
@@ -40,14 +40,18 @@ const (
 	dataFolderVolumeName            string = "/.rocketpool/data"
 
 	PruneFreeSpaceRequired uint64 = 50 * 1024 * 1024 * 1024
-	dockerImageRegex       string = ".*/(?P<image>.*):.*"
-	colorReset             string = "\033[0m"
-	colorBold              string = "\033[1m"
-	colorRed               string = "\033[31m"
-	colorYellow            string = "\033[33m"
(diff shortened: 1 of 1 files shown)


rocketpool-cli/service/service.go
```diff
@@ -40,14 +40,18 @@ const (
 	dataFolderVolumeName            string = "/.rocketpool/data"
 
 	PruneFreeSpaceRequired uint64 = 50 * 1024 * 1024 * 1024
-	dockerImageRegex       string = ".*/(?P<image>.*):.*"
```
jshufro, line 43: the `./` is removed so we capture everything before the `:`

PR #857, opened 2025-06-18
Remove cid from submissions

0xfornax:
As discussed, we have stopped using CIDs and, considering they can be the reason for missing consensus, we have decided to remove them from reward submissions.


jshufro (approved):
double check that this will work in the bindings? otherwise

PR #861, opened 2025-06-25
Change Reth prune params

0xfornax:


shared/services/rocketpool/assets/install/scripts/start-ec.sh
jshufro, line 1: can we put each param on a separate line for readability, just repeating cmd="$cmd --param 0" eg

PR #865, opened 2025-07-01
nethermind-v1.32.2

0xfornax:


jshufro:
the upgrade path needs to be tested. I'm not sure if the default 1024 is serialized, rendering this code ineffective

PR #874, opened 2025-07-11
Add Ultrasound Filtered Relay Configuration

### LINKED ISSUE #873: adding ultra sound relay to the regulated set
ultra sound relay recently started supporting proposers which insist on all their blocks being ofac filtered or nothing.

similar to titan we offer both services, so max-profit stays the same but if someone wants the filtering they need to connect to our filtering instance.

i see the config here: https://github.com/rocket-pool/smartnode/blob/0cc5252d22459a60ae0dc899ca12598acc67719a/shared/services/config/mev-boost-config.go

i'll create a PR to add a second filtered config similar to titan, let me know if i missed anything

alextes:
closes #873 

also updates the staging url to point to the new hoodi environment as the old holesky environment has mostly shut down.

### DIFF
--- a/shared/services/config/mev-boost-config.go
+++ b/shared/services/config/mev-boost-config.go
@@ -46,6 +46,9 @@ type MevBoostConfig struct {
 	// Ultra sound relay
 	UltrasoundRelay config.Parameter `yaml:"ultrasoundEnabled,omitempty"`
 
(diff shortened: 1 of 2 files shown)


0xfornax:
Thank you for the contribution!

PR #875, opened 2025-07-15
Update golang to 1.24.5

jshufro:
I really want to use some new-as-of-1.22 features in the std http package, namely, the more advanced handler routing: https://go.dev/blog/routing-enhancements

Probably time we just update to latest stable.


jshufro:
ran `make release` to confirm cross-compilation is still viable.

PR #891, opened 2025-08-08
Skip recover of inactive keys

### LINKED ISSUE #784: Wallet rebuild/recover should skip validators that belong to finalized/closed minipools

0xfornax:
Closes #784

### DIFF
--- a/shared/utils/wallet/recover-keys.go
+++ b/shared/utils/wallet/recover-keys.go
@@ -13,6 +13,7 @@ import (
 	"github.com/rocket-pool/smartnode/bindings/rocketpool"
 	"github.com/rocket-pool/smartnode/bindings/types"
 	"github.com/rocket-pool/smartnode/shared/services"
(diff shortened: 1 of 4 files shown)


shared/utils/wallet/recover-keys.go
```diff
@@ -51,6 +52,23 @@ func RecoverMinipoolKeys(c *cli.Context, rp *rocketpool.RocketPool, address comm
...
 
+	// Get validator statuses by pubkeys
+	statuses, err := bc.GetValidatorStatuses(pubkeys, nil)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting validator statuses: %w", err)
+	}
+
+	// Filter out inactive validators
+	filteredPubkeys = []types.ValidatorPubkey{}
+	for _, pubkey := range pubkeys {
```
jshufro, line 63: seems like this could be simpler with `if (withdrawn) continue`

PR #894, opened 2025-08-17
refactor: replace []byte(fmt.Sprintf) with fmt.Appendf

yinwenyu6:
Optimize code using a more modern writing style. Official support from Go, for more details visit https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.


jshufro:
can you drop the appendf call and use strconv.Quote to remove the sprintf instead? The real issue here is using *f to add double quotes to a string, not the byte slice caat.

yinwenyu6:
> can you drop the appendf call and use strconv.Quote to remove the sprintf instead? The real issue here is using *f to add double quotes to a string, not the byte slice caat.

Of course. Modified.

PR #966, opened 2025-11-28
refactor: use strings.Builder to improve performance

rifeplight:
strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)


0xfornax:
Branch has conflicts and very small improvements that won't make it worth for us to merge it.

PR #1015, opened 2026-02-25
Ipv6 stack

0xfornax:
Add IPv6 support

### DIFF
--- a/shared/services/rocketpool/assets/install/templates/exporter.tmpl
+++ b/shared/services/rocketpool/assets/install/templates/exporter.tmpl
@@ -34,3 +34,6 @@ services:
     network_mode: host
 networks:
   net:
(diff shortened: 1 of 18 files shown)


shared/services/rocketpool/assets/install/templates/exporter.tmpl
```diff
@@ -34,3 +34,6 @@ services:
     network_mode: host
 networks:
   net:
+    {{- if .IsIPv6Enabled}}
+    enable_ipv6: true
```
jshufro: any reason not to just have `enable_ipv6: {{ .IsIPv6Enabled }}` for all of these?

thomaspanf: Oh that's a good shout. Nice find!

PR #1020, opened 2026-02-28
external mode alerts

b0a7:
This PR limits some of the alerts that appear when your EC/CC are externally managed.  Critical warnings about ports that aren't open aren't applicable in an externally managed configuration.  For port warnings, the configuration is checked in the code that checks for port connectivity.

Tested via new unit tests


b0a7:
> I actually disagree with the premise here. Low disk space alerts are important, even for VC-only machines. They have a tendency to fill up with logs or docker images- and when they do fill up, clients can't write-ahead to slashing protection databases and stop attesting.
> 
> It seems to me that the `AlertEnabled_LowDiskSpaceCritical` flag can be use by anyone who wants to manually disable these alerts. The default should be on.
> 
> However, `node_filesystem_avail_bytes` is a bad metric to threshold against. The alert should default to comparing to `total_bytes` so we can fire when below, say, 15% of the drive's total space remains.

I agree it is a better idea to keep the critical disk space alerts and adjust the threshold.  To simplify things I reverted the changes about the disk space alerts and will just focus on the port checks since it is a more obvious problem & solution.

PR #1033, opened 2026-03-05
Fix treegen voting power logic for megapools

jshufro:


treegen/voting-power.go
```diff
@@ -40,7 +40,7 @@ func getNodeVotingPower(s *state.NetworkState, nodeIdx int) *big.Int {
 	// Get provided ETH (32 * minipoolCount - matched)
 	ethProvided := big.NewInt(activeMinipoolCount * 32)
 	ethProvided.Mul(ethProvided, oneEth)
-	ethProvided.Sub(ethProvided, node.EthBorrowed)
+	ethProvided.Sub(ethProvided, node.MinipoolETHBorrowed)
```
jshufro: megapool eth borrowed is accounted for just below. EthBorrowed includes megapools, but here we only want to calculate the portion due to minis

PR #1078, opened 2026-04-21
enable-QF1001

0xfornax:


jshufro:
solid readability win, right?

PR #1079, opened 2026-04-21
Bump go, update tools and fix cache permission error on build

jshufro:


0xfornax:
Should we use 1.26.2?

0xfornax (approved):

PR #1097, opened 2026-05-06
Optimize proof memory

0xfornax:
Total totalAllocDeltaBytes -47.1% 
Peak Heap (heapInusePeak) -3.9%

Reusing the state tree we obtain a ~38% improvement on the execution time while also reducing peak heap and total allocation bytes.


jshufro:
Pretty smart. Does it work?

You can also consider vectorizing the function- right now we do this one validator index at a time, but there are node operators who have to generate multiple proofs per pass.

0xfornax:
> Pretty smart. Does it work?

It works!

> You can also consider vectorizing the function- right now we do this one validator index at a time, but there are node operators who have to generate multiple proofs per pass.

The request I received was to optimize the memory footprint as some nodes were getting an OOM. Not sure if storing multiple proofs on a single pass would be an improvement on the memory issue.

shared/types/eth2/fork/electra/state_electra.go
```diff
@@ -183,6 +183,62 @@ func (state *BeaconState) ValidatorProof(index uint64) ([][]byte, error) {
 	return append(proof, blockHeaderProof.Hashes...), nil
 }
 
+// ValidatorAndSlotProof produces both the validator proof and the slot proof
+// for the state's current slot lowering the memory cost for building the proofs.
```
jshufro, line 86: This should be implemented in terms of the same two function bodies that are used when we only compute the individual proofs, otherwise you have way too much duplicated logic.

shared/types/eth2/fork/fulu/state_fulu.go
```diff
@@ -186,6 +186,62 @@ func (state *BeaconState) ValidatorProof(index uint64) ([][]byte, error) {
 	return append(proof, blockHeaderProof.Hashes...), nil
 }
 
+// ValidatorAndSlotProof produces both the validator proof and the slot proof
+// for the state's current slot
+func (state *BeaconState) ValidatorAndSlotProof(validatorIndex uint64) ([][]byte, [][]byte, error) {
```
jshufro, line 90: same applies here, of course

jshufro (changes requested):
The manual 'release' of these byte buffers to the gc is a bit of an antipattern, see https://github.com/rocket-pool/smartnode/pull/1098 for an approach that removes the need.

PR #1115, opened 2026-06-19
Make it so esc saves changes before exiting page when modifying textboxes

jshufro:
tested quickly and sloppily


0xfornax:
I'm worried about all the long time users who built the muscle memory that ESC is used when you don't want to save changes.

jshufro:
> I'm worried about all the long time users who built the muscle memory that ESC is used when you don't want to save changes.

I guess I can make it configurable.

jshufro:
> > I'm worried about all the long time users who built the muscle memory that ESC is used when you don't want to save changes.
> 
> I guess I can make it configurable.

I started doing this and no longer think it's worth it.

Just to be clear, though, the current behavior (before this PR) is that editing a textbox and pressing ESC doesn't just undo the edits to the textbox, it goes to the home page/previous page without showing you that the edits were reverted, which is a terrible UX. Either we need to adopt this PR or change it so the `ESC` key is no longer a shortcut for 'back' (tab->enter becomes the only way to exit subpages)

As an aside, config.Parameter is ripe for turning into a generic.

0xfornax (approved):

PR #1134, opened 2026-07-14
Revert same port config

0xfornax:
Teku 26.7.1’s same-port dual-stack logic ([PR #10936](https://github.com/Consensys/teku/pull/10936)) skips the IPv4 bind when ports match and :: is present. It keeps only the IPv6 wildcard as one dual-stack listener. That works on a normal host but under Docker with published [::]: ports it breaks IPv4 P2P.


yorickdowne:
https://github.com/Consensys/teku/issues/10972

PR #1141, opened 2026-07-25
Remove shared/utils

jshufro:
It is no secret that I dislike "shared/utils/..."

Every time I see it, my brain thinks "shared means stuff that has no other home, and utils means stuff that has no other home" so it just seems redundant.

Beyond that, it becomes a dumping ground for situations where developers either haven't thought critically about where to put some new code, or can't resolve some import cycle.

I have removed a lot of shared/utils to directories in rocketpool/ and rocketpool-cli/, and in a few cases where it makes sense for things to be truly 'shared', i have simply moved them up a directory (eg shared/utils/hex->shared/hex).

Eventually, we should also remove shared/services and shared/types to be closer to their respective functionality in either rocketpool/ or rocketpool-cli/, and potentially rename `shared` to `utils`. At that point it'll only have a few useful math functions, colored logging, and some hex string manipulation, which are small and dumb enough to really fit the "utils" naming.

NB that there is a change to node-collector that switches from querying the bn/ec for validator/minipool balances to getting them from the NetworkState object which will need checking up on. Everything else should be more or less non-functional.


rocketpool/node/collectors/node-collector.go
```diff
@@ -791,44 +788,39 @@ func (collector *NodeCollector) Collect(channel chan<- prometheus.Metric) {
...
-		totalDepositBalance += eth.WeiToEth(minipool.NodeDeposit)
-		totalNodeShare += eth.WeiToEth(minipool.NodeBalance)
-		totalBeaconBalance += eth.WeiToEth(minipool.TotalBalance)
+	for _, minipool := range minipools {
+		validator, exists := state.MinipoolValidatorDetails[minipool.Pubkey]
+		if exists {
+			totalBeaconBalance += math.GweiToEth(validator.Balance)
+		}
+		totalDepositBalance += math.WeiToEth(minipool.NodeDepositBalance) + math.WeiToEth(minipool.UserDepositBalance)
+		totalNodeShare += math.WeiToEth(minipool.NodeDepositBalance)
```
jshufro: See here for the change that needs closer inspection. Previously this queried the BN, now it gets all the data from the state.

rocketpool/eip712/eip712_test.go
jshufro, line 1: This functionaltiy is _so straightforward_ I was surprised it didn't already have a test.

Issue #1149, opened 2026-07-28
`wallet rebuild` context deadline exceeded


lostmsu:
> Rebuilding node validator keystores...
> Could not rebuild wallet: error calling HTTP API POST /api/wallet/rebuild: Post "http://127.0.0.1:8280/api/wallet/rebuild": context deadline exceeded

version 1.20.7

PR #1154, opened 2026-07-31
Removed the timeout from wallet recover as large nodes were having issues

0xfornax:
Wallet recover and rebuild on large nodes can exceed the CLIs default HTTP timeout. This PR drops that timeout. The daemon now allows only one recover, test-recovery, or rebuild at a time, exposes progress via GET /api/wallet/recovery-status, and rejects concurrent attempts with the in-flight operation and key progress. The CLI now confirms the effects of those commands, polls and prints recovery progress, and reports if work is already running on the daemon.


0xfornax:
Closes #1149

thomaspanf (approved):

PR #1181, opened 2026-08-18
Bump tcell dependency

0xfornax:


rocketpool-cli/service/tui_term.go
```diff
@@ -0,0 +1,48 @@
...
+	termProgram := os.Getenv("TERM_PROGRAM")
+	if isUnsupportedTUITerm(term, termProgram) {
+		return nil, unsupportedTUITermError(term, termProgram)
+	}
+
+	app := tview.NewApplication()
+	md := cliconfig.NewMainDisplay(app, oldCfg, cfg, isNew, isUpdate, isNative)
+	if err := app.Run(); err != nil {
+		return nil, unsupportedTUITermError(term, termProgram)
+	}
```
jshufro: this may misinterpret a non-term error as a term error, better to delete all the custom detection and simply compare `if err == tcell.ErrTermNotFound ...`

This PR also doesn't fix the lack of decorated error messages in configureService and configureServiceHeadless

Issue #1140, opened 2026-07-25
rocketpool/api/node/rewards/rewards.go should use networkstate filtered on the NO's minipools/megapools


jshufro:
This is a pretty old command and while it uses multicall it is still less efficient than using a NetworkState.

PR #1182, opened 2026-08-19
Use NetworkState in node rewards. 

thomaspanf:
A handful of improvements to the node rewards path: https://github.com/rocket-pool/smartnode/issues/1140
- Use NetworkState instead of multicall. 
- Scope the NetworkState to a node's megapool instead of walking the entire network for megapool data.
- Use NetworkState to compute the megapool CL rewards. Previously, we made an extra API call from the CLI to get megapool CL rewards. 
- Fixed an accounting bug with exited-unfinialized minipools: `Your share of unskimmed Beacon Chain (CL) rewards is currently -8.005268 ETH.`

Highly scientific benchmark on two different hoodi nodes: 

```
user@node:~/dev/smartnode$ time rocketpool node rewards
```
| Minipools | Megapool validators | Before | After |
|---|---|---|---|
| 2 | 26 | 4.262s | 0.723s |
| 1400 | 1705 | 13.961s | 4.572s |


0xfornax:
There seems to be a regression related to `GetHeadStateForNode`. It no longer loads the network-wide megapool set and the node daemon's supply collector still treats that slice as global, so megapool network supply metrics would be node only.

jshufro:
> There seems to be a regression related to `GetHeadStateForNode`. It no longer loads the network-wide megapool set and the node daemon's supply collector still treats that slice as global, so megapool network supply metrics would be node only.

Metrics collection should either not use the GetHeadStateForNode if it cares about network-wide aggregates, or should not care about network-wide aggregates.

Personally, I think having network stats in grafana is a bit pointless, as better third-party trackers exist (Dune, steely's site). All it does is increase load on the node every time the state locker cache thingy refreshes.

0xfornax:
> Personally, I think having network stats in grafana is a bit pointless, as better third-party trackers exist (Dune, steely's site). All it does is increase load on the node every time the state locker cache thingy refreshes.

I'm in favor of removing global stats that are impacting the SN performance. Just make sure to adapt the Grafana dashboard so users understand these changes.

0xfornax (approved):

PR #1185, opened 2026-08-20
Api improvements

0xfornax:
Adds configurable Smart Node HTTP API exposure (closed, localhost, or external) and a dedicated API settings category with a new API token. Remote and local clients must authenticate according to a token-requirement mode (all routes, or only mutating ones such as transactions, wallet operations, and validator exits).

Adds rate limit capabilities. The default will not enforce limits.

Request logs now include the client IP.


rocketpool/node/sensitive.go
```diff
@@ -0,0 +1,218 @@
+package node
+
+import "strings"
+
+// submitsTransaction is every HTTP path that submits an execution-layer
+// transaction or a consensus-layer voluntary exit. Checked first so a tx
+// route cannot be treated as read-only by prefix rules.
+var submitsTransaction = map[string]struct{}{
```
jshufro: I think the architecture here is poor. Ideally there's a `type SmartnodeRoute interface` somewhere that defines certain attributes of routes, and each route itself has a `AllowUnauthed() bool` function that returns false when a route is sensitive.

You can have the `SmartnodeRoute` interface extend [http.Handler](https://pkg.go.dev/net/http#Handler) like this:
```go
type SmartnodeRoute interface {
  http.Handler
  GetPath() string
  AllowUnauthed() bool
}
```

You will also probably benefit from using the [Middleware](https://www.alexedwards.net/blog/making-and-using-middleware) pattern in golang, and most of your routes can be a basic type like
```go
type WriteRoute struct {
  path string
  //handler http.HandlerFunc //not actually needed, just have a `func (w *WriteRoute)HandleHTTP(...)` instead
  middlewares [](func(next http.Handler) http.Handler)
}
```

This will let you add common logic across routes via middleware, and have a compile-time enforcement on what is considered safe vs unsafe, and give you a reasonably good way to have per-route-type api permissions.

As it is, a developer who adds a write route will almost certainly forget to update `submisTransaction`, and we will probably never discover that they forgot.

rocketpool/api/snroute/snroute.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"net/http"
+	"strings"
+
+	"github.com/rocket-pool/smartnode/rocketpool/api/response"
+	"github.com/rocket-pool/smartnode/shared/services/apitoken"
+)
+
+// Route is an HTTP handler bound to a path with a compile-time auth class.
+type Route interface {
```
jshufro, line 19: consider adding
```golang
func (route Route) RegisterTo(router Router) {
    router.Handle(route)
}
```

so externally you can simplify the routes.go files with code like this
```golang
func RegisterRoutes(router *snroute.Router, c *cli,Command) {
    // Read handlers
    snroute.Read("/some/url", someURLHandlerFunc).RegisterTo(router))
    // Write handlers
    snroute.Write("/some/other/url", someOtherURLHandlerFunc).RegisterTo(router)
}
```

It's nice to have a 1-line-per-route registration, then the body of these functions can stay closer to their source. That is to say, instead of:
```golang
	router.Handle(snroute.Read("/api/auction/can-claim-lot", func(w http.ResponseWriter, r *http.Request) {
		lotIndex, err := parseLotIndex(r)
		if err != nil {
			response.WriteErrorResponse(w, err)
			return
		}
		resp, err := canClaimFromLot(c, lotIndex)
		response.WriteResponse(w, resp, err)
	}))
```
you have
```golang
    snroute.Read("/api/auction/can-claim-lot", canClaimLotHandler).RegisterTo(router)
```
in routes.go and
```golang
func canClaimLotHandler(w http.ResponseWriter, r *http.Request) {
		lotIndex, err := parseLotIndex(r)
		if err != nil {
			response.WriteErrorResponse(w, err)
			return
		}
		resp, err := canClaimFromLot(c, lotIndex)
		response.WriteResponse(w, resp, err)
}
```
in claim-lot.go, where it "naturally belongs"

OK if you want to merge without this refactor though, this is still very much a step in the right direction

shared/services/apitoken/token.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+const (
+	Prefix     = "rpsn_"
+	rawBytes   = 32
+	fileMode   = 0600
+	dirMode    = 0700
+	tokenBytes = 64 // hex-encoded rawBytes
+)
+
```
jshufro, line 16: You should seriously consider adding types
```golang
type ApiToken [32]byte
type ApiTokenFile struct {
    Token ApiToken `json:"token"`
    Comment string `json:"comment"`
}

func (a ApiToken) MarshallText() []byte {
    // returns the rpsn_-prefixed string
}
```
and persisting json files with user-prompted comments, so people can easily remember why they made tokens and revoke them as needed.

This also makes the token file format easier to migrate to new formats in the future, since you can attempt to parse the json in multiple ways.

shared/services/apitoken/token.go
```diff
@@ -0,0 +1,177 @@
...
+	buf := make([]byte, rawBytes)
+	if _, err := rand.Read(buf); err != nil {
+		return "", fmt.Errorf("could not generate API token: %w", err)
+	}
+	return Prefix + hex.EncodeToString(buf), nil
+}
+
+// Equal compares two tokens in constant time. Length mismatches still run a
+// dummy compare so the timing does not leak the expected length.
+func Equal(a, b string) bool {
```
jshufro: If you take the above suggestion this just becomes a 
```golang
func (a *ApiToken) Equals(other *ApiToken)
```

and you can do the same for Valid() etc

shared/services/config/api-config.go
```diff
@@ -0,0 +1,159 @@
...
+	ApiPort config.Parameter `yaml:"apiPort,omitempty"`
+
+	// How the API port is published
+	OpenApiPort config.Parameter `yaml:"openApiPort,omitempty"`
+
+	// Bearer token (stored in a sidecar file)
+	APIToken config.Parameter `yaml:"-"`
+
+	// Which routes require the bearer token
+	TokenScope config.Parameter `yaml:"tokenScope,omitempty"`
```
jshufro, line 47: Scoping is usually done on a per-token basis. As it is, you have a single api token which becomes a shared secret between the operator of smartnode and the consumer of the token.

It may be reasonable for an operator to want to issue multiple tokens, eg, a write-scoped token to a trusted application that they themselves maintain, and a read-scoped one to an application that they operate but don't control and may susceptible to supply chain attacks.

You could fix this by having the file be a `tokens.json` file with a list of tokens and their respective scopes, or you could look into something a bit more elegant like HMAC (which has trade-offs, eg, it becomes harder to revoke a token once issued, but the token scope can become a part of the token itself instead of being persisted).

Regardless of what you decide, i recommend moving the scope out of the config. It would be weird to set read-only scope and then change it after the fact to include write, since you elevate the trust level of everything that uses the token.

shared/services/config/rocket-pool-config.go
```diff
@@ -1365,8 +1375,49 @@ func (cfg *RocketPoolConfig) GetECStopSignal() (string, error) {
...
+		return ""
+	}
+	port := cfg.Api.ApiPort.Value.(uint16)
+	return fmt.Sprintf("\"%s\"", portMode.DockerPortMapping(port))
+}
+
+// SyncAPITokenFromDisk loads the API token from the sidecar file, creating one
+// if needed. forCLI selects the host data-dir path; the daemon uses the
+// in-container path.
+func (cfg *RocketPoolConfig) SyncAPITokenFromDisk(forCLI bool) error {
```
jshufro: the `forCli` or `isNative` mode thing really bugs me, maybe on startup we can just add a bool to the `cfg` object that records whether it's cli or api, and another that reports native vs not, and then remove the argument from the function?

jshufro (approved):
A few suggestions but this feels like a good step forward overall so nothing urgent or blocking

shared/services/config/api-config.go
0xfornax: Accepted the read/write scoped tokens suggestion but kept a server option to still accept read requests without a token.

PR #1200, opened 2026-08-30
Separate on-chain state from reconstructed data

jshufro:
Previous, NetworkState had a mix of fields: fields that directly map to storage on either the evm or beacon chain mixed with fields derived from the former.

This PR separates the derived fields into a separate, encapsulating type. This is a small reduction in tech debt, as maintainers no longer need to worry about whether or not they have called `CalculateAverageFeeAndDistributorShares` before accessing the state.

Further deconstruction of the treegen state machines will make it easier to make them more modular in the future. We update them to taking the state index from NewGenerator to GenerateTree, which shifts the dependency on data away from init-time and towards runtime. Ultimately we should fully remove state from the treegen structs and pass data to its subroutines on a case by case basis.


0xfornax (changes requested):
GetBulkMegapoolDetails is being called with no addresses and returns an empty map

0xfornax (approved):

PR #1204, opened 2026-09-01
Add proof bundles

0xfornax:
Moves megapool beacon proofs onto versioned, ABI-encoded proof bundles so Smart Node matches BeaconStateVerifier on v1.4.1-dev / Gloas.

Stake, notify-exit, not-exit, dissolve-with-proof, and notify final balance now take `(proofVersion, proofData)` instead of separate proof structs. Call sites encode `ValidatorProofBundleV1` (or final-balance v1/v2) before submitting. Final-balance v2 also packs next-withdrawal-index and validator-balance proofs.

### DIFF
--- a/bindings/megapool/proof-bundle.go
+++ b/bindings/megapool/proof-bundle.go
@@ -1,0 +1,132 @@
+package megapool
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+// Proof versions accepted by BeaconStateVerifier on v1.4.1-dev.
+var (
+	ValidatorProofVersion1    = big.NewInt(1)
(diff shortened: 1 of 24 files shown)


bindings/megapool/proof-bundle.go
```diff
@@ -0,0 +1,132 @@
...
+	SlotProof       SlotProof
+}
+
+type FinalBalanceProofBundleV2 struct {
+	WithdrawalProof                  WithdrawalProof
+	ValidatorProof                   ValidatorProof
+	SlotProof                        SlotProof
+	PreviousNextWithdrawalIndexProof NextWithdrawalIndexProof
+	ValidatorBalanceProof            ValidatorBalanceProof
+}
```
jshufro, line 34: Should these types have json tags? They seem like useful artifacts to be able to serialize on the network or to disk.

PR #1214, opened 2026-09-08
 Pin Docker images digests 

### LINKED ISSUE #360: [Minor Security Risk] Depending on external dockerhub images
# TL;DR
It looks like rocket pool has a variety of docker images that are pulled in without full digest/sha256 checks.

# Scenario 1
- A rocket pool team member is compromised and docker hub is used to push a new image overriding the existing images.
- Some users will begin using new images that could steal their node keys.

# Scenario 2
Currently rocket pool uses optional external images like here:
https://github.com/rocket-pool/smartnode/blob/b8b63459678317fd635d08e93384c9c101a505ca/shared/services/config/prometheus-config.go#L8
- A prometheus developer is compromised and replaces the `v2.43.0` tag with a malicious image
- Rocket pool nodes may now be at risk of this malicious code (like stealing keys).

# Additional Comments
The way docker-compose file is setup it should not be at risk because it does not expose sensitive data to the `prometheus` container, however, it is a bad idea to assume malicious code cannot exit containers. For example a few years ago there was the `libc` vulnerability and clever ways of breakouts happen all the time.

For example, some of the containers rocket pool deploys mounts the `docker.sock`, which is well known to be a vulnerability (see: https://redfoxsec.com/blog/introduction-to-docker-container-breakouts/).

I don't expect rocket pool to be 100% secure from every case, but blindly trusting external images seems risky.

# Proposal
The simplest way to solve this is to also tag the digest. This can be problematic, since docker does not support multiple digests for multiple architectures and rocket pool supports multiple architectures. We could have a map in the code of the digests associated with each platform, sniff out the platform we are currently running on and then replace the magic `EXPORTER_CONTAINER_TAG` environ with one that has the digest matching the platform's digest.

Another idea is upon service start/restart via cli, we first query images that are downloaded and verify that the digest is one of the confirmed white listed images. In the event that it is not, we prompt the user and give a way to override for some advanced users before starting the image(s).

0xfornax:
This PR aims to harden Smart Node security by pinning the image digests of all officially released images. Once pinned, Smart Node will not pick up a newly published malicious image, even if it uses the same tag.

Image digests are stored in `images.env` as `repo:tag@sha256:<index-digest>`. Docker verifies the digest and still picks the right architecture from the multi-arch index or fails if it the digest has changed. TUI tag edits write to the same file and are replaced by the official ones on the next install.

Closes #360


shared/services/rocketpool/client.go
```diff
@@ -1167,18 +1179,102 @@ func (c *Client) compose(composeFiles []string, args string) (string, error) {
...
+			return fmt.Errorf("could not write %s: %w", path, err)
+		}
+	}
+	return nil
+}
+
+// writeComposeMain writes runtime/compose.yml with include + env_file so catalog
+// interpolation happens in Compose. Later env files override earlier ones
+// (mainnet.env then {network}.env). TUI overrides are set in the process environment.
+func writeComposeMain(composeMain, projectDir string, envFiles []string, deployed []string) error {
```
jshufro: I'd prefer to see txt/template population from a `compose.tmpl` so that an `overrides/compose.yaml` file can be used to do global overrides. Not a huge deal, but I find using golang string building logic to create valid yaml to be a bit fragile

One technique I like to use with templated config is to create functions like:
```
func TestnetOnly(cfg) {
if cfg.IsTestnet() return "" else return "#"
}
```

Then in the template you can use `{{.TestnetOnly}}- testnet.env` or similar to include overrides for testnet.

.gitignore
```diff
@@ -2,6 +2,7 @@
 vendor
 node_modules
 /scripts/config.js
+/scripts/pin-images.sh
```
jshufro: seems like maybe `/scripts/*` is warranted instead of individual filenames, since you seem to have a whole directory of non-revisioned scripts

shared/services/rocketpool/client.go
```diff
@@ -1167,18 +1179,102 @@ func (c *Client) compose(composeFiles []string, args string) (string, error) {
...
+
+	composeFileFlags := []string{
+		fmt.Sprintf("-f %s", shellescape.Quote(composeMain)),
 	}
 	for _, container := range composeFiles {
 		composeFileFlags = append(composeFileFlags, fmt.Sprintf("-f %s", shellescape.Quote(container)))
 	}
 
-	// Return command
-	return fmt.Sprintf("COMPOSE_PROJECT_NAME=%s docker compose --project-directory %s %s %s", cfg.Smartnode.ProjectName.Value.(string), shellescape.Quote(expandedConfigPath), strings.Join(composeFileFlags, " "), args), nil
```
jshufro, line 1180: these %s command formatters always drove me crazy, thank you for killing this

shared/services/rocketpool/client.go
```diff
@@ -1167,18 +1179,102 @@ func (c *Client) compose(composeFiles []string, args string) (string, error) {
...
+	// Compose interpolates the process environment ahead of include env_file.
+	envPrefix := []string{
+		fmt.Sprintf("COMPOSE_PROJECT_NAME=%s", cfg.Smartnode.ProjectName.Value.(string)),
+	}
+	overrides := cfg.ComposeEnvOverrides()
```
jshufro: since we're using a receive on `cfg` it would be sensible to put the string building there. e.g
```
cfg.ComposeEnvOverridesCmdFormat() []string-> "foo=bar", "baz=buz"
```

It simplifies the logic here, which is not overly concerned with the values, just the result of the double-join (`" ".join("=".Join(k, v))`)

shared/services/rocketpool/client.go
```diff
@@ -1167,18 +1179,102 @@ func (c *Client) compose(composeFiles []string, args string) (string, error) {
...
+	overrideKeys := make([]string, 0, len(overrides))
+	for key := range overrides {
+		overrideKeys = append(overrideKeys, key)
+	}
+	sort.Strings(overrideKeys)
```
jshufro: I assume the rational behind sorting is sound, but it's hard to tell without a comment. Also a good reason to move this into cfg where you can add a test that asserts sortedness

jshufro (commented):
This looks good to me now and I think it's a real improvement.

Since we're parsing the .env files to float logic into the tui, it might make sense to update the TUI to display something like "The default version for this service is sigp.io/lighthouse:v1" when an override is present, or "This is the default version of this service for the current smart node release" when not

shared/services/rocketpool/client.go
jshufro: oh and you need to test that a space character present in a env var value or name causes an error

shared/services/rocketpool/client.go
0xfornax: The sort is there for easier diffs

jshufro: That's a good reason, just put a comment in the code so we don't have to guess

shared/services/rocketpool/client.go
0xfornax: OK. In practice it got a little more complicated with
```
env_file:
      - {{$.MainnetEnv}}
      {{$.TestnetOnly}}- {{$.TestnetEnv}}
      {{$.DevnetOnly}}- {{$.DevnetEnv}}
{{- range $.ExtraNetworks }}
      {{$.CommentUnless .}}- {{$.EnvFile .}}
``

PR #1227, opened 2026-09-16
Stop with the old binary

0xfornax:
Using the old binary to stop so we don't have errors from unexpected missing resources.


rocketpool-cli/update/update.go
```diff
@@ -109,6 +109,13 @@ func Update(yes bool, skipSignatureVerification bool, force bool) error {
 	}
 	fmt.Printf("Replacing the cli at %s with the latest version...\n", oldBinaryPath)
 
+	backupPath := oldBinaryPath + ".bak"
+	cpCmd := exec.Command("cp", oldBinaryPath, backupPath)
```
jshufro, line 113: is there maybe an os.Copy instead?

rocketpool-cli/update/update.go
```diff
@@ -109,6 +109,13 @@ func Update(yes bool, skipSignatureVerification bool, force bool) error {
...
+	if err := cpCmd.Run(); err != nil {
+		return fmt.Errorf("error backing up current cli: %w", err)
+	}
+	defer os.Remove(backupPath)
```
jshufro: err is unchecked, might want to wrap it in a func that prints a warning to stderr

jshufro (approved):

